PR #831, opened 2014-05-29
Channel fix - check max page size

timhobbs:
Page size can be 500, but some channels cannot take that, so this checks
the feature result to get the max page size and adjusts on the fly.

### DIFF
--- a/MediaBrowser.WebDashboard/dashboard-ui/scripts/channelitems.js
+++ b/MediaBrowser.WebDashboard/dashboard-ui/scripts/channelitems.js
@@ -115,54 +115,61 @@
         }
 
         query.folderId = folderId;
-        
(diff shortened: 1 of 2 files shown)


MediaBrowser.WebDashboard/dashboard-ui/scripts/channelitems.js
```diff
@@ -115,54 +115,61 @@
         }
 
         query.folderId = folderId;
-        
-        $.getJSON(ApiClient.getUrl("Channels/" + channelId + "/Items", query)).done(function (result) {
 
-            // Scroll back up so they can see the results from the beginning
-            $(document).scrollTop(0);
+        $.getJSON(ApiClient.getUrl("Channels/" + channelId + "/Features", query)).done(function (result) {
```
timhobbs, line 119: Not sure if you know of another way to do this, but since it is async I could only think to nest the calls.

PR #933, opened 2014-10-07
Adding the 192kbps bitrate to link the new 144p resolution now available...

jscoys:
... into the webclient


LukePulverenti:
merged manually.

PR #957, opened 2014-12-22
Fix authentication bug for usernames with special characters

Ceten:
Add html decoding to allow for usernames with special characters when
authenticating by username. Clients need to be adapted to encode the username too if they want to be able to authenticate usernames with special characters. If clients don't html encode the username, this works just as before.


LukePulverenti:
This is no longer needed given the new username requirements.

PR #968, opened 2015-01-07
Ssl in mediabrowser against new listener.

jabbera:
This is actually all that is required to get ssl working against core mediabrowser and the new socket listener. I'm not a web guy, so before I get into that I'd like any and all comments against the core changes.

Big issues I see: 

How do specify the certificate both programatically and in the UI. (Location of he pfx is hard coded in socketlistener.)

How to redirect MBC to the ssl port, not the http port.
Provide a way to disable http?
Do we need 2 pubic ports (http[s])?
Any places a url is generated with http hard coded.
Anything else anyone can think of?


LukePulverenti:
For starters, re-work the certificate so that the path is passed in. yes we'll need two ports, and i don't think we need to worry about disabling http yet. probably too early for that. 

if you can resubmit against the dev branch that would be great.  i'll merge it in as soon as the current beta goes stable.

PR #969, opened 2015-01-08
Ssl in mediabrowser against new listener.

jabbera:
Can't change pull request target branch apparently. (Or I couldn't figure out how.)

This has been updated to allow certificate to be configurable, although it's hard coded right now. Will update the UI once the internals have been decided on\approved.

Whats the UDP port used for?

### DIFF
--- a/MediaBrowser.Controller/IServerApplicationHost.cs
+++ b/MediaBrowser.Controller/IServerApplicationHost.cs
@@ -37,6 +37,28 @@ namespace MediaBrowser.Controller
         int HttpServerPort { get; }
 
         /// <summary>
+        /// Gets the HTTPS server port.
+        /// </summary>
+        /// <value>The HTTPS server port.</value>
+        int HttpsServerPort { get; }
+
+        /// <summary>
+        /// Gets the value indiciating if an https port should be hosted.
+        /// </summary>
+        /// <value>
+        /// The value indiciating if an https port should be hosted.
+        /// </value>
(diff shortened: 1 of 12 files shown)


jabbera:
What are the advantages of using the native web socket listener instead of SocketHttpListener listener? While ssl is certainly supported the netsh hoops you need to jump through make it PAINFUL:-)

LukePulverenti:
you don't have to worry about that. it will be deprecated over the next couple of releases.

jabbera:
Great.

jabbera:
HTML changes. I have no idea what I'm doing. Just copy and pasted some related fields.

LukePulverenti:
I made some changes prior to the stable release to get the https port property added and get the ball rolling on getting the port authorized for all new users via the initial admin prompt. You'll need to update this to use that so that the pull request can be merged in automatically.

Also, deprecating the WebSocketPortNumber property is good, but we can't do it yet because it will likely break the iOS app. So that needs to be restored temporarily, just have it point to the http port.

jabbera:
Rebased. Can't test from where I am though.

LukePulverenti:
need to update again and re-submit (sorry). once you do that i'll try to merge it in this weekend so we can start testing. thanks.

LukePulverenti:
also just for consistency can you rename CertificateLocation to CertificatePath in all places? thanks.

jabbera:
All set.

PR #984, opened 2015-01-17
Show ssl port on dashboard

jabbera:
I've added ssl port information to the dashboard. I only added the label to the en_US file. I'll be happy to add it in any other file. 

First: is there an automated process for this that I don't know about? 
Second, if the language isn't English do I add the english label anyway or just leave the label out?

You are probably going to get a bunch of small pulls from me at this point. I get precious little time to do any "fun" work with a job, wife, and child. Let me know if they get too granular.


LukePulverenti:
Actually you only need to do it in the javascript.json file. our translation site will take care of the language specific ones the next time i sync it.

jabbera:
Fantastic. This should be better the.

PR #986, opened 2015-01-19
Create self signed cert if one does not exist

jabbera:
This will auto generate a self signed certificate in windows if one does not exist. Most of the code was borrowed from an msdn blog. Question:

Is INetworkManager an appropriate interface to add this functionality to.

I'm too lazy to implement a linux implementation. I don't have an easily accessible VM and adding a reference to Mono.Security seems like a headache. (Which version of mono versus which Mono.Security)


LukePulverenti:
not really, and putting it on the interface suggests that all implementations can support it. I see there's a mono.security nuget package, so that would be preferable. but it seems old (dec 2013).

jabbera:
That was my concern about about mono.security as well. There is a version with each mono build and it seems as if that package hasn't been updated in a while. Unless reflection is broken out implementing this on linux will be a PITA. 

I'm not sure where else to put the code so feel free to suggest somewhere else, or not merge it.

LukePulverenti:
this looks easy enough, no? https://blog.lextudio.com/2014/05/how-to-create-certificates-in-c-via-mono-security/

we will just need our own build of Mono.Security

jabbera:
Mono.Security is HUGE. Apparently it's standalone so I've got a certificate generation code working. Just making sure it doesn't fall over on a modern version of mono. 

I hate developing for linux:-)

jabbera:
Updated to include mono cert generation. I've included Mono.security as a third party library but it SHOULD NOT be included in the install.  We just need it to compile against. All installations of mono include their own Mono.Security. As long as they don't make a breaking change we are good. I took the Mono.Security from Mono 3.2.8 and ran it against 3.12.0 it works fine.

LukePulverenti:
ok, i slightly reworked the hosting options so you'll have to re-submit. I renamed UseHttps to EnableHttps, and also changed the SystemInfo reporting to not reflect configuration but instead the current state. Also made sure the automatic port mapping respects the setting as well as immediately reporting to Connect after a change.

Will Mono.Security work on windows running under the .net framework? I'm guessing  yes since it's a nuget package. We already use Mono.NAT in a similar fashion.

Also wondering if we should have one public port mapped to either local port depending on settings, or have two public ports. Right now it is a single port. I am thinking of the scenario where an admin wants to require https but allow an exception where needed if for whatever reason https isnt' possiblew.

jabbera:
Will rebase shortly.

Will Mono.Security work on windows running under the .net framework? 
  It may. I'm not really sure. I didn't use the nuget package since it was so old. I implemented the code twice, once using pInvoke and once using Mono.Security. Investigating the this will require some time, and the weekend is over so I'll be short on it till next week. If this is a requirement it will have to wait.

Also wondering if we should have one public port mapped to either local port depending on settings, or have two public ports.

It's probably a good idea until all the 3rd party clients are updated though. That way new ones can use https and old ones will still work. If it's one port and only one, many clients will not work properly I'm guessing. Especially if you don't purchase a valid certificate from a provider. There needs to be an option to disable the http public port though for people who only want encrypted traffic.

LukePulverenti:
Ah my mistake, i missed that the p/invokes were still there. Ok. I'll add the second port then and this should be almost good to go.

So I am guessing the field to specify a cert will only be for those who want to use their own? Are you keeping it as an override?

jabbera:
That controls where the cert is located. If the server starts up and no cert is located at that path, it will generate one. If one is there it will leave it be. Basically it has to be set for ssl to work. I couldn't think of a reasonable default that would work on linux and windows. If one exists it probably makes sense to just default the field.

jabbera:
Rebased.

LukePulverenti:
does this require elevation?

jabbera:
Not as long as the cert file is in a location that doesn't require elevation

jabbera:
New ui organization looks nice.

neagix:
This type of implicit behavior feature should have been at least disabled by default, as of now it is impossible to start Emby without SSL, because the check in `ApplicationHost.cs` about CertificatePath being empty always returns false (albeit it is truly empty in configuration).

To make things worse, `HttpsEnabled` seems to be checked only by UI and not by the actual code that starts the HTTP server.

PR #991, opened 2015-01-27
QuickSync updates and NVENC support

mjb2000:
Introduce support for NVENC.
Correct handling of levels for QS and NVENC
Correct rounding of h*w for QS


mjb2000:
Whoops! - I've made a mistake here....

PR #995, opened 2015-01-31
Bluebull32 patch 1

Bluebull32:
BIF related fixes:
1) Give ffmpeg more time to create thumbnails, but periodically check to make sure the process did not hang.
2) Reverse Endianness for integer values in the index.bif file.

### DIFF
--- a/MediaBrowser.Api/Playback/BifService.cs
+++ b/MediaBrowser.Api/Playback/BifService.cs
@@ -155,7 +155,7 @@ namespace MediaBrowser.Api.Playback
         private byte[] GetBytes(int value)
         {
             byte[] bytes = BitConverter.GetBytes(value);
-            if (BitConverter.IsLittleEndian)
+            if (!BitConverter.IsLittleEndian)
                 Array.Reverse(bytes);
             return bytes;
         }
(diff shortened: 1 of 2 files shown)


LukePulverenti:
Thanks! Just so you're aware, I don't intend to always have the BIF generated on the fly. As you can tell, it's much too slow. Ultimately I intend to have options to create it ahead of time during scheduled tasks and/or at the point of media discovery. If we can confirm that this proof of concept results in a usable BIF, then I'll go ahead and put that work on my plate for the near future.

Bluebull32:
Great, thanks.  I completely agree that the .bif files would be a lot more useful if they were generated ahead of time.  If you are going to look at this later in testing, be aware the the bif files will not get picked up during playback when they are ready.  The video needs to be stopped, and resumed/started over before the thumbnails will be displayed.

The stuff on the roku client app is good to go, it just need to be uncommented.

LukePulverenti:
Thanks. So you've run a successful test from the Roku?

Bluebull32:
Yes I have.  I just keep an eye on the the folder where the thumbnails and index.bif are created.  Once the index.bif file is created, I can just replay the video on the roku, and the thumbnails function just like you would expect.

Now that I think of it, there is probably no reason to keep the jpegs around after the index.bif has been successfully created.

PR #1000, opened 2015-02-05
TV Guide starts at current time and is based on 1 minute increments

Techywarrior:
This is an update to pull request #998 that includes starting the guide at the current time instead of the beginning of the current hour (and therefore showing potentially ended shows)


LukePulverenti:
Ok, cool. can you close these and resubmit on the dev branch? thanks.

PR #1002, opened 2015-02-06
TV Guide starts at current time and is based on 1 minute increments

Techywarrior:
TV guide works much more like a real TV guide now with the start time of the current day being the start of the guide instead of the beginning of the hour of the current day (which could show ended programs). Also increments of 1 minute are used so show cells reflect actual start/end times in respect to time header. Also will solve issue of short shows not appearing in the guide if two shows were in the same 30 minute block (like two 15 minute shows back to back where one starts at :00 and the other at :15)

### DIFF
--- a/MediaBrowser.WebDashboard/dashboard-ui/css/livetv.css
+++ b/MediaBrowser.WebDashboard/dashboard-ui/css/livetv.css
@@ -202,7 +202,8 @@
     overflow-y: hidden;
     overflow-x: scroll;
     float: left;
-    border-bottom: 4px solid #3B3B3B;
+	/*border-bottom: 4px solid #3B3B3B;*/
+	padding-bottom: 4px;
 }
 
 .programGrid {
--- a/MediaBrowser.WebDashboard/dashboard-ui/css/livetv.css
+++ b/MediaBrowser.WebDashboard/dashboard-ui/css/livetv.css
@@ -158,7 +158,6 @@
 .timeslotCell {
     display: inline-block;
     border-bottom: 1px solid #444;
-    border-left: 1px solid #444;
     border-collapse: collapse;
     position: relative;
 }
@@ -216,8 +215,11 @@
     margin-left: 190px;
 }
 
-.timeslotCell, .timeslotHeader {
-    width: 190px;
+.timeslotHeader {
+    width: 179px;
+}
+.timeslotCell {
+	width: 6px;/* (180/30 min) */
 }
 
 .timeslotCellInner {
@@ -233,6 +235,7 @@
 }
 
 .timeslotCellInnerWithProgram {
+	border-left: 1px solid #444;
     z-index: 99;
 }
 
--- a/MediaBrowser.WebDashboard/dashboard-ui/scripts/livetvguide.js
+++ b/MediaBrowser.WebDashboard/dashboard-ui/scripts/livetvguide.js
@@ -63,587 +63,582 @@
 
         channelsPromise.done(function (channelsResult) {
 
-            ApiClient.getLiveTvPrograms({
-                UserId: Dashboard.getCurrentUserId(),
-                MaxStartDate: nextDay.toISOString(),
-                MinEndDate: date.toISOString(),
-                channelIds: channelsResult.Items.map(function (c) {
-                    return c.Id;
-                }).join(',')
+			ApiClient.getLiveTvPrograms({
+				UserId: Dashboard.getCurrentUserId(),
+				MaxStartDate: nextDay.toISOString(),
+				MinEndDate: date.toISOString(),
+				channelIds: channelsResult.Items.map(function (c) {
+					return c.Id;
+				}).join(',')
 
-            }).done(function (programsResult) {
+			}).done(function (programsResult) {
 
-                renderGuide(page, date, channelsResult.Items, programsResult.Items);
-                hideLoadingMessage(page);
+					renderGuide(page, date, channelsResult.Items, programsResult.Items);
+					hideLoadingMessage(page);
 
-            });
+				});
 
-            var channelPagingHtml = LibraryBrowser.getPagingHtml(channelQuery, channelsResult.TotalRecordCount, false, [10, 20, 30, 50, 100]);
-            $('.channelPaging', page).html(channelPagingHtml).trigger('create');
+			var channelPagingHtml = LibraryBrowser.getPagingHtml(channelQuery, channelsResult.TotalRecordCount, false, [10, 20, 30, 50, 100]);
+			$('.channelPaging', page).html(channelPagingHtml).trigger('create');
 
-            $('.btnNextPage', page).on('click', function () {
-                channelQuery.StartIndex += channelQuery.Limit;
-                reloadChannels(page);
-            });
+			$('.btnNextPage', page).on('click', function () {
+				channelQuery.StartIndex += channelQuery.Limit;
+				reloadChannels(page);
+			});
 
-            $('.btnPreviousPage', page).on('click', function () {
-                channelQuery.StartIndex -= channelQuery.Limit;
-                reloadChannels(page);
-            });
+			$('.btnPreviousPage', page).on('click', function () {
+				channelQuery.StartIndex -= channelQuery.Limit;
+				reloadChannels(page);
+			});
 
-            $('.selectPageSize', page).on('change', function () {
-                channelQuery.Limit = parseInt(this.value);
-                channelQuery.StartIndex = 0;
-                reloadChannels(page);
-            });
-        });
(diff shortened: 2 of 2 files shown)


tadams88:
Just tested this pr on my Nexus 4. I timed how long it took to render the guide after the page loaded (starting when the popup appears, and ending when the popup disappears).

dev: 1s, 1s, 1s
pull-request: 21s, 21s, 20s

Decreasing the cell size from 30m to 1m seems to have a very negative impact on mobile performance. This might need some performance tuning.

I do think it is a good idea to allow programs to be displayed accurately to-the-minute. I think we would get better performance from re-working the code to iterate through the programs once, and place each program in its appropriate spot, rather than iterating through the cells and then iterating through the programs once for each cell to find the correct program to put in that cell.

I do agree that the starting time should not be the beginning of the hour, but using current time means that the currently playing program is completely illegible for the last 15 minutes of the show. Personally I would rather see the most recent half-hour, instead (e.g. 1:00, 1:30, 2:00, etc.). I think this is how the guide works for Comcast at least.

Techywarrior:
Yea, I was worried about the performance. On desktop you don't even notice it but the number of new DIVs was a concern. I think it's the rendering that's causing the lag and not the actual looping. I didn't rewrite the entire guide, I just "fixed" it so that shows actually reflected when they started/ended and some shows that were hidden would be displayed.

I wasn't as concerned about the shows being hard to read right at their end because you can always hover over them and see the titles. Showing the current half hour may work but it would still potentially display shows that have ended (if it was only 15 minutes long). I think the better solution may be to display it sort of how I did but stretch the width of the current partial 30 min block to be the same width as 30 minutes. There would be a bunch of things to consider with how it will work however.

But, someone else can do all that. Since there was absolutely zero feedback from the people who could merge this either approving it or asking for specific changes or fixes I'm not going to bother spending time to rewrite the entire guide code to support a more robust solution.

LukePulverenti:
Apologies I haven't been able to loom at it. I asked him to take on the review for it so that it could get done sooner rather than later.

Techywarrior:
Ok, well if it's actually going to be reviewed and (ultimately approved) then I can spend the time rewriting the entire thing to be more efficient.

LukePulverenti:
Yes definitely. I am wondering if 5 or 10 min separation could be a quick fix that will avoid a refactor. Keeping in mind nothing should change visually from the original dressing except for the presence of the shows that are missing

Techywarrior:
I did some thinking and testing and I don't think that doing exactly what tadams88 is suggesting will work. There is a chance that no show is scheduled in a particular timeslot. Not entirely sure if I should even worry about that but if we looped thru the programs and there was a gap between two there would be a problem.

I rethought the old method that was used and I think I have come up with some good changes. First I removed all the individual DIVs used for "time slots" and am using the programs to fill those spaces themselves and then creating a filler DIVs if there isn't a program scheduled (if it ever happens). This reduces a single channel's DIV count from 1440 to the number of programs (~40-50). I also realized that I was being stupid checking every minute. If a program is running nothing else can be on at the same time so it now won't check until the time the current program actually ends. This change reduced the number of times I looped thru the programs from 6160 to 152 (on my 7 channel system).

I'll try get some more people to test and confirm that everything behaves as expected and then update the pull request.

tadams88:
What I was suggesting would be a major overhaul of how the guide is rendered. The only reason is because that's how I tackled it when building a proof-of-concept guide overlay for the video screen. It involves calculating and setting the size of the program container, and then positioning each program absolutely based on calculations of where it should go. It seemed to be a pretty performant approach, but I agree that if we can get the functionality improvements we are gunning for without such a major overhaul, that would be the better option at this point.

I didn't realize that we still had a div for every minute, sitting underneath each show. That should be a very helpful fix for div count. And good move on skipping slots that are already filled. That will cut WAY down on iterations through the programs array. Love it.

I wonder if the program iterations would be even more efficient if after we found the program (around line 175) we added ```programs.splice(i, 1);``` That would mean each iteration gets shorter, and saves us from iterating through programs that have already been placed. The only question there is whether the cost of calling ```programs.splice()``` once per program found would be more or less than the cost of iterating through already found programs. My gut says the .splice() would add efficiency but might be worth a perf test.

Or, if we are sticking with 1 minute increments, I wonder if there would be a performance gain from turning the array in to an object with date.setSeconds(0) as the key.
```javascript
var programDict = {};
programs.filter(function (curr) {
        return curr.ChannelId == channel.Id;
    }).forEach(function(curr) {
        try {
            var time = parseISO8601Date(program.StartDate, { toLocal: true })
                                .setSeconds(0).getTime().toString();
            programDict[time] = curr;
        }
        catch (err) {

        }
    });
```
Then the programs could be retrieved by calling ```programDict[date.getTime().toString()]``` instead of iterating through an array?

Dunno, just bouncing ideas around.

Techywarrior:
My gut feeling is that having everything absolutely positioned would create additional overhead in browser rendering.

I could do what you suggest with removing each program but not sure it is going to make a big difference. It's looping thru only the programs for the specific channel at a time so the array is only going to be 50-60 items at max. More likely it will be a lot less. On top of the splice I'd have to create a new global variable (which shouldn't effect too much).

Let me know what sort of performance you get with the new update compared to the last test you did and we can take it from there (updated files are on the forum. I didn't want to keep making pull requests here to test. Or PM me if there is a better way to work/test this)

tadams88:
What you just sent me in PM is a massive performance improvement. Great work. Back down to ~1s to render the guide on my Nexus 4. (A lot better than 20s!)

If we are at 1sec render time we probably don't need to worry overmuch about tweaking out every last bit of performance. So just ignore my rambling musings :)

One last thing would be what you mentioned with stretching the first cell out to 30min width, regardless of how long it is.

This is how WMC shows the guide at 4:59, for instance
wmcguide

Having a 1min sliver makes it look at-a-glance like the guide didn't render correctly, just in my personal opinion
mbsguide

If you can figure that one out, I think this would be an awesome improvement to the guide!

Techywarrior:
Good news, testing looks good with partially populated channels or channels with no data. I also stretched the first partial time block the same as WMC does it.

(I'm sorry about all the lines where just the spacing changed but I couldn't fix it to match no matter what I did)

Techywarrior:
I've removed the last functionality that would cause 1000s of DIVs to be created. Now if there is no program at a specific time, or no guide data, I am finding the next program on that channel and making the spacer DIV large enough to fill all the space. Worst case scenario this replaces 1440 DIVs with 1 DIV at the expense of having to loop thru the list of programs for a given channel one time each time there is no guide data.

LukePulverenti:
We also need the original formatting preserved on the javascript because right now the diff indicates that nearly every single line in the file has changed, so you can't really tell what's what.

LukePulverenti:
Also comments in the css - for example, why do borders and paddings need to be changed when the issue is missing data.

Techywarrior:
I can't tell what sort of spacing you are using for the indentation. I tried tab and spaces and neither matched up with the existing code. Shouldn't Github ignore spaces?

I don't know how I can apply all my changes to a clean copy and not have the same issue with spacing. 

The CSS had to be fixed because the way it was setup before was that every 30 minutes was a block in the guide, that DIV had a border on it regardless of whether the program actually ended there. Then the program DIV was on top of that and hid the borders in the intermediary DIVs. This makes no sense so I changed it to have the border on the actual program DIV. The display is the same but it's more efficient.

I also changed the width of 30 minutes to be evenly divisible by 30 so I could have a fixed width for "1 minute".

Techywarrior:
All the spacing has been fixed. It should be much easier to read the diff now.

LukePulverenti:
The current release attempts to resolve this with a quick fix by changing the cell duration to 10 minutes down from 30. I am guessing the issue not solved but rather has just been improved to occur slightly less often. We can always drop it down to 5. But this is the kind of surgical precision change I was hoping for:

https://github.com/MediaBrowser/MediaBrowser/commit/908a059f4512cdf06d736b9d201beaba6d47f1f3#diff-f2e3b21e04d0097edffc0982d3cd3317

Techywarrior:
Changing it down to 10 minutes will make it happen less but still won't fix the problem. Shows end at odd times not just on evenly divisible by 5/10 times. From looking at the TV guides of people in Europe it seemed even more common for shows to NOT end on a 5/0 then to end on one. 

I understand that simply changing the cell time to 10 minutes is simpler but my pull request fixes a lot of the issues with the TV guide besides just that. We've tested it pretty well and I got rid of almost anything that would slow it down.

LukePulverenti:
It's all done. I've rewritten the rendering. There are no more timeslots now, just a blank canvas with programs on it. So the issue is solved now.

PR #1014, opened 2015-02-18
Update TvShowsService.cs

Bluebull32:
Fix for 25 episode limit on Roku app:  "TotalRecordCount" in line 497 needs to be total episodes in season, not the number of episodes returned in this query response.


LukePulverenti:
will merge manually into dev. thanks.

PR #1026, opened 2015-02-25
Prototype of max task length for scheduled tasks.

jabbera:
Here is a prototype of what I am thinking for capping task runtime. There are a handful of issues to discuss.

Should the task runtime be a property of the task or the trigger? Right now it's just hard coded for a sample. Making it part of the task seems natural except it would not be possible to make the runtime variable based upon the time of day. I don't know if we care about this.

The task needs to be smart enough to know how to make progress each time the task is activated. If the task always takes an hour, but is only allocated 30 minutes it will always do the first 50%. For BIF creation this shouldn't be an issue since it will pick up where it left off. What about other tasks that may want a time cap?

Can I modify IScheduledTask or ITaskTrigger? Are these public APIs? How do plugins use them? I usually use base classes for these types of thing with an interface so you can add members with sensible defaults, but existing binaries continue to work.

### DIFF
--- a/MediaBrowser.Api/ScheduledTasks/ScheduledTaskService.cs
+++ b/MediaBrowser.Api/ScheduledTasks/ScheduledTaskService.cs
@@ -195,7 +195,7 @@ namespace MediaBrowser.Api.ScheduledTasks
                 throw new ResourceNotFoundException("Task not found");
             }
 
-            TaskManager.Execute(task);
+            TaskManager.Execute(task, null);
         }
 
         /// <summary>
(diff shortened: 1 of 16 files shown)


ebr11:
Personally, I think the time limit should be a part of the trigger because you may want the same task to run for different lengths of time at different times (say five hours during the night but all day on Mondays).

LukePulverenti:
Yes the worker should do what the trigger tells it. What I would do is add a TaskExecutionOptions object as a parameter to the Execute method. For now, give it one property, say MaxRuntimeMs. 

This property can be set based on trigger options - example, a strict time limit, or via some of the situations Ebr mentioned above.

Then in ExecuteInternal to enforce the time limit, it's trivial because I'm pretty sure there's a constructor overload for CancellationToken to cancel after a given duraiton.

jabbera:
This turns out to be very difficult due to the design of the triggers and the task manager. I've left some comments in the recent checkin. I'll need some suggestions.

jabbera:
Never mind. I figured it out.

jabbera:
This is MUCH more involved then the previous implementation. Let me know what you think before I flush it out.

LukePulverenti:
Yes its' a little messy now that I see it. Let me think it over. Options on the task itself might be cleaner.

PR #1028, opened 2015-02-26
Task based approach to delay

jabbera:
Here is an alternative approach. The majority of the edits are due to the lack of base class for IScheduleTask.


LukePulverenti:
No, that's out. IScheduledTask only needs to worry about execution. Options come from the user, not the developer of the task.

jabbera:
Bummer.

PR #1029, opened 2015-02-26
Prototype 3 of max task length for scheduled tasks.

jabbera:
This is a little more palatable by using default params on the interface. (Could use overloads as well, personal choice.)


jabbera:
Any more thoughts on this? There are a bunch of people coming up with odd workarounds because task length can't be limited:

http://mediabrowser.tv/community/index.php?/topic/18822-recursive-batch-create-roku-bif-files/?hl=bif
http://mediabrowser.tv/community/index.php?/topic/18755-create-bif-file-manually/?hl=bif

LukePulverenti:
yea this one looks good. i'll take a closer look later and merge.

PR #1031, opened 2015-03-02
Throttle with priority instead of pausing threads.

jabbera:
I encourage you to look at using process priority to throttle instead of using thread suspension. There is no reason to stop transcoding when you hit the "limit" if there is spare CPU cycles. The current change will effectively waste idle CPU cycles. (Assuming the whole video will be watched, even if you don't it just deleted anyway) If you continue to transcode when the CPU is idle the next new video request that comes in may not have anything to compete with.

The implementation should work on linux\windows with the same API so there is only one ProcessManager implementation to maintain. You could probably ditch it all together and just do the work in the transcoding throttler. 

It also requires no changes to ffmpeg.

Note: Proof of concept. Untested.


LukePulverenti:
Thanks. I'm going to keep the same api but switch the base implementation to what you're doing with Idle/Resume.

PR #1034, opened 2015-03-07
Transcode segments

Bluebull32:
Fix for client app hanging at the end of playback for transcoded item. (http://mediabrowser.tv/community/index.php?/topic/8001-media-browser-for-roku-channel/page-27#entry182857)

In theory, there may be a small speed performance hit in the transcoding process, but I did not observe any slower speed during my testing. This was done to fix a problem specific to the Roku, but could possibly be an improvement for other apps.  I only tested Roku.

In the existing implementation: When transcoding begins, the server provides that client with playlist of segments.  This is a prediction of the segment names in order and the running time for each segment, because the segments have not been created yet.  When the segments actually do get created, the segment lengths (in seconds) do not exactly match the predictions.  More importantly, the differences in segment lengths causes the number of actual segments to be less than the prediction in many cases.

When the roku app nears the end of the playback, it will request the last segment, n (it was told that this segment would exist by the prediction in the beginning).  This segment does not exist, because the actual transcoding only produced n-1 segments.  The roku app waits for this segment to become available, which it never does, so the app hangs indefinitely.

This modification changes the command line options for ffmpeg when creating segments.  The segments created from these options more strictly adhere to the segment length requested, and the actual number of segments will match the prediction.  The roku will get the last segment when it requests it, and can finish the playback without a problem.


LukePulverenti:
Awesome, thanks. I will test and merge on dev.

ebr11:
Yes, Hallelujah as this will fix this problem on the ATV app as well.

PR #1035, opened 2015-03-11
Preferences should dismiss when you hit 'OK'

### LINKED ISSUE #882: Preferences should dismiss when you hit 'OK'
The new user preferences screen has OK and Cancel buttons on it but, when you hit OK it just saves the changes and then sits there.  Meaning the user has to either hit Cancel or back to get out of the screen.

Based on the labeling, I would expect OK to dismiss the screen and go back to where I came from.

tadams88:


LukePulverenti:
I don't think this one's a good idea now that there are four tabs in the preferences. if you want to change settings in all four you keep getting bounced back to the home page after each one. I don't think that's good.

tadams88:
I see your point. I do agree with ebr that the current buttons are odd though. I have to click OK, and then click 'cancel' to accept my changes. There should be a way to 'positively' close the preferences page.

Other ideas:
* Old-school windows interface would be to have an 'Apply' button that has current functionality, as well as an 'Ok' button that saves, and closes the preference window
* Add apply buttons per section, and change "ok" button at the bottom to have text "Save and close". Quick mock: 
applyheader

One other thought: Because the "ok" button is at the bottom of the page, do you really think people scroll to the bottom of the page to hit ok, just so they can scroll back up and change a setting they just scrolled past? I feel like the natural flow would be to change settings as you scroll down, and once you reach the button click "save and close"

LukePulverenti:
you don't have to click cancel to get out, you can just navigate. i think you're only thinking that way because you as a developer happens to know that cancel will do a history.back().  i think a save button in the top right would be a bit of an uncommon paradigm, and inconsistent with how the rest of the web interface works. in it's current form i wouldn't say that it's difficult to use so my preference is to leave it as it is. i haven't heard people asking how do they return to the home screen from inside user preferences.

tadams88:
Ah. I think the issue is that the ok button makes me think of this as a modal rather than a page. Maybe change ok to 'save'? Either way, perhaps close issue #882

LukePulverenti:
Yea that will work

PR #1038, opened 2015-03-14
Add playlist next/previous buttons to video player

tadams88:


tadams88:
Desktop Full Screen:
desktop_full

Desktop 556px:
desktop_556px

Desktop 555px:
desktop_555px

Mobile:
mobile

Personally I think it might be a good idea to float the controls left in mobile, as they are somewhat separate in function from the buttons they are with.

PR #1040, opened 2015-03-16
Expose HTTPS

DirtyJerz:
HTTPS is noop without this.


DirtyJerz:
Should I also make this change under MediaBrowser.build?

PR #1041, opened 2015-03-18
More logical page load on item delete

Techywarrior:
On delete, remove item from jsTree and load sibling or parent depending on existence of prev/next sibling.


LukePulverenti:
Merging manually into dev.

PR #1046, opened 2015-03-20
Accept the formats {hostname} and {hostname}/ and http://{hostname} 

jose-pr:
If the url supplied to the client didn't contain the protocol or the trailing / the server would return a System.ArgumentOutOfRangeException.
Allows more valid formats.


jose-pr:
Added a method to insert additional headers with HttpRequestOptions

LukePulverenti:
Merged manually but all requests should be properly prefixed by developers. Mono behaves a little differently than .NET and will sometimes throw exceptions over this.

PR #1055, opened 2015-03-27
Recognize images by extension

jose-pr:
Some time the server providing the image treat the data as binary added
an option to recognize that is an image by the format extension.


timhobbs:
It would probably be more efficient (no string concatentation) to include the "." in the list: ".png", ".jpg".

LukePulverenti:
Well this was added because when a program in Next Pvr doesn't have an image, it return a 200 status code instead of 404 along with a mime type that is not expected for an image. So I look at this pull request and i say that's why we have mime types in the first place.

LukePulverenti:
no longer needed.

PR #1063, opened 2015-04-02
Enable IsAnamorphic to function properly

paul-777:
Allow IsAnamorphic and related transcoding for dlna devices that can't handle anamorphic videos to function properly as discussed here
http://emby.media/community/index.php?/topic/18905-isanamorphic-not-working-right/


LukePulverenti:
can you rework this to use the resolution normalizer? because clients specify their max values and the server determines the output based on that. for example, client might say max resolution of 1080p with a max bitrate of 2mbps. The bitrate would normally be higher but the user lowered it for network reasons. At 2mbps h264, 1080p is too high so the resolution needs to be lowered. The client still indicates 1080p because that's what it supports, but the server adjusts based on the other params.

paul-777:
I have reworked it - turns out the issue was that we weren't detecting anamorphic removal as upscaling so we were passing in the current video bitrate as the limit and many of my 720 res anamorphic 21:9 videos have bitrates less than 800Kbps so were then being scaled down to 640 rather than upscaled properly. For example a 720x432 anamorphic video which should be upscaled to 1032x432 was being down sized to 640x268.

LukePulverenti:
Great. The other situation that needs to be tested to make sure it hasn't been broken is when we don't know the resolution of the source. So try live tv for example.

paul-777:
Unfortunately I don't have that capability to be able to test it.

LukePulverenti:
can you try it on some channel content? you'll need to pick one that doesn't report any media info. twitch should work.

tadams88:
Shouldn't this also be going against the dev branch?

paul-777:
Channels? All I have is the media on my own storage. I can't see why the changes would affect media where we can't obtain mediainfo though - if the IsAnamorphic flag isn't set it the metadata then it will have no effect. Which in itself will mean that those anamorphic streams (many live TV broadcast are) still won't be upscaled properly, but won't be any worse than the current situation.

paul-777:
Hi, I have now updated this again with the latest pull from master, please can you take a look at merging it - having to pull changes in from master and in some cases remake them in new files is very time consuming. This fix is important for anyone using LG (and other) TVs which cannot handle anamorphic videos. thanks

paul-777:
Updated from MediaBrowser master again

LukePulverenti:
Since we have to put out another bug fix release, i'm going to manually add the code to detect anamporhic video. so that means you'll have to update from master one more time. then once we finally get into a longer release cycle i'll merge the rest.

paul-777:
ok, no problem, thanks

paul-777:
Merged from master

paul-777:
pull from master

LukePulverenti:
Based on this I think the anamorphic detection will have to be reverted:

http://emby.media/community/index.php?/topic/21445-direct-stream-question

(Unless you can solve that). The ffmpeg params for resizing anamorphic content can stay though.

paul-777:
The issues for this should now be fixed and I have updated to master again. For it to work properly with only content the metadata does have to be refreshed to perform detection again. Many of the changes are also not to do with the detection but also dnla clients passing a flag for whether they support anamorphic to get the correct stream.

paul-777:
Hi Luke,  I've merged this with master again, is there any reason you can't merge it? Keen to get back to being able to use standard releases but need this working on all my LG TVs.
thanks

PR #1072, opened 2015-04-10
Don't use DirectoryNotFound exception for control flow.

jabbera:
This reduces the time to scan a 25k photo collection from

13 minute(s) and 18 seconds

To

1 minute(s) and 31 seconds


jabbera:
This is no longer necessary after subsequent changes by luke.

PR #1123, opened 2015-06-26
Add Emby to all webpage titles

andjscott:
What is thought about adding "Emby | " to the page titles?

My motivation is so that I can control Emby using a Macs native media keys through [beardedspice](https://github.com/beardedspice/beardedspice/). Beardedspice can find valid tabs either by [checking their URL](https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/SoundCloudStrategy.m#L24) or [their title](https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/SubsonicStrategy.m#L24). Emby's URL could be checked (specifically checking for ":8096") but the port can change.

This change doesn't change all pages but what do you think of the idea?


LukePulverenti:
Sorry there's just not enough space to do this.

PR #1125, opened 2015-06-30
Connect WebSocket connect through Reverse Proxy

BaukeFahner:
While debugging fonts and cross site stuff through reverse https:// proxy, I discovered that the WebSocket connection was unable to connect (301). Adding a slash to the base URL enabled the usage of WebSockets. I'm not sure this works for direct connections (maybe others can test?).

I couldn't find a fix in the proxy :(, this works for me. Perhaps it helps many others too. I can change my build script, but this seems the most persistent solution.

Regards,

Bauke

### DIFF
--- a/MediaBrowser.WebDashboard/dashboard-ui/thirdparty/apiclient/apiclient.js
+++ b/MediaBrowser.WebDashboard/dashboard-ui/thirdparty/apiclient/apiclient.js
@@ -398,7 +398,7 @@
             }
 
             var url = serverAddress.replace('http', 'ws');
-            url += "?api_key=" + accessToken;
+            url += "/?api_key=" + accessToken;
             url += "&deviceId=" + deviceId;
 
             webSocket = new WebSocket(url);


LukePulverenti:
Is this really the only way? this is going to require review in every single app if that is the case.

BaukeFahner:
Not sure, that is why I put it here, for review with others :). From what I gather is that the variable url does not contain a trailing slash. Added to that, you get an url that immediately gets the '?'. In other words, my reverse URL contained: https://host/emby? Where it should be https://host/emby/?. 

I tried removing the trailing / in the location part of the nginx configuration, but then the reversehost part disappears alltogether:

-- working, allowing the url https://host/emby/web/index.html to work --
        location /emby/ {
                ...
        }

-- not working, removing the emby part of my URL --
        location /emby {
                ...
        }

It could be that the base url of emby has an issue with sitting behind and reverse proxy altogether. Perhaps someone has an idea on how to pinpoint that part?

LukePulverenti:
This will be handled differently.

PR #1154, opened 2015-08-15
Handle symlinks in library folders

arcticwaters:
Here is a first attempt at handling symlinks. I've only had a chance to test this on linux, but it should work just as well with junction points and symlinks in Windows if it didn't already. I'd almost argue that this is a bug in mono, but this fix can be used for now.

Symlinks will now work just like regular files. Emby also appears to handle broken symlinks well. Broken video files will trigger metadata download, but will refuse to play with file not found errors in the logs.

Things I haven't tested for yet:
* windows
* symlink for root folders
* symlink to parent which contains the root folder
* symlink to parent which multiple user roots (in the case of different media types)

### DIFF
--- a/MediaBrowser.Api/EnvironmentService.cs
+++ b/MediaBrowser.Api/EnvironmentService.cs
@@ -221,7 +221,9 @@ namespace MediaBrowser.Api
         /// <returns>IEnumerable{FileSystemEntryInfo}.</returns>
         private IEnumerable<FileSystemEntryInfo> GetFileSystemEntries(GetDirectoryContents request)
         {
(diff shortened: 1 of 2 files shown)


LukePulverenti:
Great stuff! The extra IO call is unfortunate, but if that's what's needed then so be it. Does that replace all of our uses of that method?

arcticwaters:
There is one more in `DeleteLogFileTask.cs`. I doubt it is a very big use case to use symlinks in the logs folder.

LukePulverenti:
Ok, thanks. If you want the pull request then please close and resubmit on the dev branch. Otherwise I can just close and make the code changes. Thanks.

arcticwaters:
Closing to rebase on top of dev branch.

PR #1158, opened 2015-08-19
Do #1157 properly with relative urls.

MeltedLux:
This does the same as https://github.com/MediaBrowser/Emby.Web/commit/ff76be45e06db7ea87e604b5d1dc4f580d5e6d01

Let.
The.
Browser.
Decide.


LukePulverenti:
Is this causing a problem? We need support for the file:// protocol so I might not be able to merge this. Thanks.

psxlover:
Isn't this only in the server's web page? I don't think you'll need to run it from the file system, unless if it gets replaced by Emby.Web in which case this file would be replaced too.

PR #1159, opened 2015-08-24
Add Kodi DLNA profile, which direct plays everything

ljani:
As a side note, you might want to update Fody to at least version 1.26.2, because the current 1.19.1 doesn't play nice with Visual Studio 2015 at all.

EDIT: would it make sense to have a similar wild card for subtitle formats as there is for DirectPlayProfile codec types? Meaning Kodi could support every subtitle format with the external method.


LukePulverenti:
Thanks! I will close and merge manually into the dev branch. By the way, the .cs version isn't really needed. It's just so that I can easily dump out new versions of the built-in profiles in the event of scheme changes. Since your .cs version isn't a complete representation of the xml, it's not really needed.

ljani:
Thank you too. I was trying to slip this into master as it seemed a minor change ;) [`CONTRIBUTING.md`](https://github.com/MediaBrowser/Emby/blob/32d463d814eaa2f69156bb2f53f0759abd48ae7d/CONTRIBUTING.md) could mention about the `dev` branch.

Yeah, I was wondering what's the point of having both, `.cs` and `.xml`. I figured [you used the `.cs` files to generate the `.xml` files by hand](https://github.com/MediaBrowser/Emby/blob/32d463d814eaa2f69156bb2f53f0759abd48ae7d/MediaBrowser.Dlna/DlnaManager.cs#L517). So, I did that too. In other words, the `.xml` file is generated from the `.cs` file, and thus it should be a complete representation (because of the inheritance), if I'm not wrong?

LukePulverenti:
My mistake. I will add the .cs file.

PR #1162, opened 2015-08-31
Add QSV decoding support to ffmpeg transcoders

babgvant:


HazCod:
So this will add hardware acceleration if the intel CPU supports quicksync? Awesome!

babgvant:
CPU detection has not been added yet. I would like to add that in the future.

SickrageDK:
Will this be merge anytime soon? :)

LukePulverenti:
I will need to review more closely. Correct me if I'm wrong, but I don't think we want to be offloading everything to the GPU as I don't think it's equipped to handle parallel processing. Therefore the system will need to be smart about when to do it and when not to.

babgvant:
QS can handle multiple streams without issue. Obviously there is a limit, so at some point you'll hit < real time, but it's pretty safe to think about it the same way you think about CPU decoding limitations.

LukePulverenti:
Ok, my mistake. I see this is just decoding, not encoding.

babgvant:
Someone (you?) had already added most of the plumbing for encoding so I didn't touch that part (IIRC).
It might be necessary to make changes to fully enable encoding support in the future, but ffmpeg has issues ATM with QSV encoding so it's hard to test.

LukePulverenti:
I'm merging this in by hand, just starting with playback and we can always move it to the lower level encoding classes later for sync conversions.

Can you try to find a platform independent way to detect qsv support? Maybe the best way is to try to do something with ffmpeg and see if it spits out an error?

LukePulverenti:
Well I've added it, but I guess zeranoe does not include these decoders with their windows builds? That would be a shame.

Also in order to safeguard against users creating problems, we're going to query ffmpeg for decoder support at startup so that we can fallback to default if the user's ffmpeg doesn't have it.

babgvant:
Did you include the LiveTV probe in the manual merge? This is important because Emby assumes all live tv is MPEG2 when, at least here in the UK, is often not.

PR #1175, opened 2015-09-07
Add ability to restart the windows service

mford1:
Ok, this is my first time using Git and I obviously did something wrong - it's showing many more changes to the files than I actually made.

In any case, I added the ability for the service to be able to restart itself.

### DIFF
--- a/MediaBrowser.ServerApplication/MainStartup.cs
+++ b/MediaBrowser.ServerApplication/MainStartup.cs
@@ -28,6 +28,8 @@ namespace MediaBrowser.ServerApplication
 
         private static bool _isRunningAsService = false;
 
+        private static bool _isPowerShellInstalled = false;
+
         /// <summary>
         /// Defines the entry point of the application.
         /// </summary>
@@ -35,6 +37,9 @@ namespace MediaBrowser.ServerApplication
         {
             var options = new StartupOptions();
             _isRunningAsService = options.ContainsOption("-service");
+
+            if (_isRunningAsService)
+                _isPowerShellInstalled = CanLoadPowerShell();
 
             var currentProcess = Process.GetCurrentProcess();
 
@@ -174,7 +179,7 @@ namespace MediaBrowser.ServerApplication
         {
             get
             {
-                return !_isRunningAsService;
+                return !_isRunningAsService || _isPowerShellInstalled;
             }
         }
 
@@ -186,7 +191,7 @@ namespace MediaBrowser.ServerApplication
         {
             get
             {
-                return !_isRunningAsService;
+                return !_isRunningAsService || _isPowerShellInstalled;
             }
         }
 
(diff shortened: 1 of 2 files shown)


LukePulverenti:
Thanks. I need to be able to analyze the changes though. You just have to make sure formatting, tabs, spaces, etc are preserved. We use visual studio at default settings.

mford1:
I think maybe the problem is I was using Tortoise Git.  I'll try again later.

mford1:
Ok I tried several times but it just isn't working.  From what I'm seeing your files are not using the windows convention CR+LF but instead just have LF's in them.  When I check them in with CR+LF it's showing everything is different.

What am I doing wrong here?

psxlover:
The only program that doesn't work with LF is notepad. Projects using LF instead of CR+LF is normal, don't assume everyone works on Windows.

That absolute path in MediaBrowser.ServerApplication.csproj is a bad idea. What if a developer is using a 32bit windows? There is no "Program Files (x86)" on his system. Or if his primary drive is not "C"?

Since you just wanted to change your last commit you should have used amend instead of revert. To get rid of all those commits, and end up with only one you should squash your commits. If you are using git from a shell use:
git reset --soft HEAD~7
git commit
If you are using TortoiseGit go to the log (right click --> TortoiseGit --> Show log)  --> select the commits --> right click and choose "Combine to one commit".
After that you will have to push using force (-f).

Also you should probably have used the dev branch instead of the master.

mford1:
Thanks for the tips, I've never used git before so there's a bit of a learning curve.

The HintPath is just that, a hint.  It doesn't have to exist, it's really going to pull it from the GAC if it can't find it there.  This is just what's added by VS when you add a reference, I didn't add it manually or anything.

I don't assume everyone works on Windows, but I do, and that's what my tooling did for me.  If I need to change my default settings in VS to always output LF only, then so be it, but I would like to know that up front.

I'm using Git from the built-in MS provider for VS, so I don't see anything called a "squash" available.  Instructions on how to do this from the github site would be welcome!

I'll commit to the dev branch next time, I didn't know the proper procedures here.

mford1:
I think I updated the commit properly now (thanks for the TortoiseGit instructions).  Let me know if not.

psxlover:
> Thanks for the tips, I've never used git before so there's a bit of a learning curve.

Glad I could help :). Git is hard to master but it's a great tool.

> The HintPath is just that, a hint. It doesn't have to exist, it's really going to pull it from the GAC if it can't find it there. This is just what's added by VS when you add a reference, I didn't add it manually or anything.

You should probably change it to something like "%programfiles(x86)%\Reference Assemblies\Microsoft\WindowsPowerShell\3.0\System.Management.Automation.dll" or remove it completely, but since I'm not sure you'll have to wait for Luke's input on this.

> I don't assume everyone works on Windows, but I do, and that's what my tooling did for me. If I need to change my default settings in VS to always output LF only, then so be it, but I would like to know that up front.

Not sure if you can do something about it from VS, but at least git has an option to change LF to CR+LF when checking out and CR+LF to LF when commiting. It's "core.autocrlf" in git, I think it's the "AutoCrlf Convert" option in "TortoiseGit --> Settings --> Git --> Auto  CrLf Convert".

> I'm using Git from the built-in MS provider for VS, so I don't see anything called a "squash" available. Instructions on how to do this from the github site would be welcome!

There is no way to do it from the site, and I'm not using VS so I can't help you with that one.

> I'll commit to the dev branch next time, I didn't know the proper procedures here.

Since the dev branch has all the commits from the master branch, you can just change the base branch when comparing changes, right before making the pull request. Unfortunately as far as I know there is no way to change the targeted branch after making a pull request in Github other than closing it and making a new one.

> I think I updated the commit properly now (thanks for the TortoiseGit instructions). Let me know if not.

It looks fine. If you want to "amend" a commit in the future, I believe TortoiseGit has a checkbox for it when you go to make a commit.

LukePulverenti:
What advantage does restarting with powershell offer? We had a similar restart working a while ago I think just via command line, but we ran into the problem of the service user account not always having access to starting services.

mford1:
I wasn't aware of that version, I just noticed that it didn't allow restarting the service and had been like that for quite some time so I thought I'd try to contribute.  I thought there might have been some issue restarting via command line and knew about the PS Restart-Service command.

Under what situations does the LocalSystem account not have the ability to start/stop services?  Who/what is modifying the default?  

One advantage of using PS would be that you can build up a script internally to execute rather than having to interact via command line or batch files.  You can certainly determine if the LocalSystem account has access to start/stop services before enabling the button if that's what you want, but I'm still curious why/how it was disabled for some people?

ebr11:
The server doesn't necessarily run under the local system account because it may need access to network resources - in which case it needs to run under specific credentials to do so.

mford1:
Ok, so by default it installs under the local system account, but the cases where it was failing was because the user changed it to be a different user then?

It's possible to add permissions to the service for a specific (non-admin) user to be able to restart it, I can work on that.

LukePulverenti:
Ok. Also, I don't think we need to do the actual restart with powershell. There are ways to do it without creating that dependency.

PR #1176, opened 2015-09-09
Allow all "bonus" features always

AshleyYakeley:
The product is always considered "registered", giving a more pleasant user experience out-of-the-box. This should also reduce load on your validation server.


AshleyYakeley:
I discuss this PR here: https://www.reddit.com/r/opensource/comments/3kfcn6/forking_to_enable_subscriberonly_features_in_emby/

PR #1179, opened 2015-09-13
Add ability to restart the windows service.

mford1:
Change default service permissions so that it can be restarted by the interactive user or by the service user account.


LukePulverenti:
Can you add links to documentation for those permission strings? thanks.

mford1:
MSDN has one, but the best description I found was here:

http://blogs.msmvps.com/erikr/2007/09/26/set-permissions-on-a-specific-service-windows/

LukePulverenti:
Ok, please add to the code via comments so that others can refer to it in the future.

mford1:
Do contributors get supporter keys?  :)

PR #1181, opened 2015-09-15
Fix ffprobe keyframe detection on linux

hmflash:
ffprobe output parsing implementation only works with windows style line endings.

When running the library scan on a linux machine, ffprobe uses only newlines so the regex didn't actually split the ffprobe output.

The subsequent linq code that splits each line on '=' and calls to ToDictionary() would end up throwing an ArgumentException due to duplicate keys.

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -353,7 +353,7 @@ namespace MediaBrowser.MediaEncoding.Encoder
                 {
                     var text = await reader.ReadToEndAsync().ConfigureAwait(false);
 
-                    var lines = StringHelper.RegexSplit(text, "\r\n");
+                    var lines = StringHelper.RegexSplit(text, "[\r\n]+");
                     foreach (var line in lines)
                     {
                         if (string.IsNullOrWhiteSpace(line))


psxlover:
Shouldn't that be [\r\n|\n]?

hmflash:
No, I expect it to split on any combination of \r or \n that occurs one or more times.

psxlover:
You are right, I seem to have mistaken [] for ().

LukePulverenti:
have you also tested on windows?

hmflash:
I don't have a window environment to test on, but this unit test passed:
```
public void TestStringSplit()
{
	const string win = "one\r\ntwo\r\nthree\r\n\r\n\r\n";
	const string lin = "one\ntwo\nthree\n\n\n";

	var winSplit = Regex.Split(win, "[\r\n]+", RegexOptions.IgnoreCase);
	var linSplit = Regex.Split(lin, "[\r\n]+", RegexOptions.IgnoreCase);

	Assert.That(winSplit, Is.EqualTo(new[] { "one", "two", "three", "" }));
	Assert.That(winSplit, Is.EqualTo(linSplit));
}
```

LukePulverenti:
Thanks. Please resubmit this on the dev branch, or I can just make the code change manually.

PR #1183, opened 2015-09-19
Auto-Organize: Perform leftover-file and empty-folder deletion in processed folders only

softworkz:

### DIFF
--- a/MediaBrowser.Server.Implementations/FileOrganization/TvFolderOrganizer.cs
+++ b/MediaBrowser.Server.Implementations/FileOrganization/TvFolderOrganizer.cs
@@ -98,35 +98,14 @@ namespace MediaBrowser.Server.Implementations.FileOrganization
 
                 if (options.DeleteEmptyFolders)
                 {
-                    foreach (var subfolder in GetDirectories(path).ToList())
+                    if (!IsWatchFolder(path, watchLocations))
                     {
-                        DeleteEmptyFolders(subfolder);
+                        DeleteEmptyFolders(path);
                     }
                 }
             }
 
             progress.Report(100);
-        }
-
-        /// <summary>
-        /// Gets the directories.
-        /// </summary>
-        /// <param name="path">The path.</param>
-        /// <returns>IEnumerable{System.String}.</returns>
-        private IEnumerable<string> GetDirectories(string path)
-        {
-            try
-            {
-                return Directory
-                    .EnumerateDirectories(path, "*", SearchOption.TopDirectoryOnly)
-                    .ToList();
-            }
-            catch (IOException ex)
-            {
-                _logger.ErrorException("Error getting files from {0}", ex, path);
-
-                return new List<string>();
-            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


softworkz:
Sorry for the commit/revert noise. I am new to GIT (but not to development).

ebr11:
Might I suggest a Dictionary instead of a List so that you don't re-process dups...?

psxlover:
You can "squash" all those commits into one https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History,
although looking in your commit history you just need to throw away the last for commits:
```
git reset --hard HEAD~4 
```
(Watch out before using this, you'll "lose" those four commits, if you want to keep the changes from the second commit you should probably first make a new branch)
And then
```
git push -f
```
to force push to the server and overwrite the commits there.

Also next time when you want to work on two different features for pull requests you should create a branch for each.

LukePulverenti:
i thought your first implementation was fine. it was just a three line change. we don't need a setting for this.

softworkz:
@psxlover: Hmpf, yeah, now I realized that I need to create a branch for each feature and that I need to keep my forked main branch clean. Thanks for the commands I will try it to get my main branch clean.

@Luke: The setting is not for "Perform leftover-file and empty-folder deletion in processed folders only"
It is for a different feature "Disable multi-episode detection" which I mixed up because I didn't know that I need to create separate branches.

@ebr11: These file operations are pretty fast, so from a performance perspective it wouldn't really matter, but I agree that using a dictionary or even better HashSet<string> should be used instead

softworkz:
OK, I have finally managed to move the second feature to a separate branch and created another pull request for that. Thanks to @psxlover I also managed to get rid of the unnecessary commits!

softworkz:
Finallly, I have changed the List to a HashSet as proposed by @ebr11 
(and even managed to merge it into a single commit using git rebase - I'm learning... ;-)

LukePulverenti:
One thing I see just from eyeballing - the delete empty folders option is no longer going to work. Given a processed folder A, it will delete all empty sub-folders under A, but not A itself if is empty.

softworkz:
Excellent point. I fixed it now while also making sure that actual watch folders do not get deleted in case a media file was found there

LukePulverenti:
Please add comments in the code about why Path.GetFullPath is necessary, and also please use string.equals(a,b,OrginalIgnoreCase)  like we do elsewhere.

softworkz:
Casing is not the only aspect in which two file paths pointing at the same target could differ:

- 8.3 vs full naming
- Environment variables
- Symbolic links
- Navigation components (C:\temp\..\temp\.\ == c:\temp)

are the issues I wanted to rule out.

softworkz:
But it turns out that GetFullPath resolves 8.3 naming and indirections (\..\) while it does not resolve environment variables and casing.
After some research: A definite solution is not available using .NET and would require p/Invoke (GetFileInformationByHandle). 
I left GetFullPath in the commit but added comment and StringComparison.OrdinalIgnoreCase as requested.

softworkz:
From the discussion, one might get the impression that  this change request is questionable. But in fact, the discussion was just about reliable folder comparison, and the comparison method used in this change is probably slightly better than the usual employed method with no drawbacks involved. 
So I think this is still a good change...

LukePulverenti:
If you resolve the pull request conflicts i'll merge this in.

softworkz:
I don't understand why it says conflict. There is no conflict in the changd file..

softworkz:
I created a new pull request which does not believe in a conflict: #1199

PR #1185, opened 2015-09-19
Re-designed item identity providers

TomGillen:
This is a new, greatly simplified, attempt at separating item identification from metadata collection (#811). This pull request mostly consists of deleting the old code.

Rather than have special "identity" types for every media type, the providers now place their identity information into ProviderIds as a string.

e.g. the TVDB episode identifier adds `{ "TVDB-Full": "seriesid:seasonindex:episode" }` to the episode's ProviderIds. This ID is then used later when fetching metadata.

The previous implementation made the assumption that every provider would want to store information of the same format for a particular item type. This is a poor assumption, and prevented me making proper use of it in the Anime plugin - which was the original motivation in the first place.


LukePulverenti:
Excellent, much simpler!

PR #1186, opened 2015-09-20
Auto-Organize: Added new option "Disable multi-episode detection"

softworkz:
Default behaviour remains iunchanged.
Once the option is activated, the ending episode number is always
ignored, but could still be set manually in the organize dialog


softworkz:
The original purpose of this change was to overcome problems with multi-episode detection.

When the changes from 

https://github.com/MediaBrowser/Emby/pull/1189 and https://github.com/MediaBrowser/Emby.Naming/pull/4

are incorporated, then there is probably no more need for this change!

PR #1189, opened 2015-09-21
Auto-Organize: Fix PathTooLongException due to long EpisodeTitle

softworkz:
Especially with Multi-Episodes, the EpisodeTitle can become quite long,
since all comprised titles get concatenated into it.
By default, the EpisodeTitle is included in the single and multi-episode
naming patterns and this in turn can quickly lead to a
PathTooLongException.
This fix tries to keep as much as possible from the title string while
keeping the final path within the allowed limit.


LukePulverenti:
Why the param changes on GetEpisodeFileName? is that actually needed?

softworkz:
It is required to supply the length of the season (=target) folder path, because this length is required to calculate the max filename length.

hmflash:
Why not just prefix the paths with "\\?\" and not be limited by MAX_PATH?

https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx
http://blogs.msdn.com/b/bclteam/archive/2007/02/13/long-paths-in-net-part-1-of-3-kim-hamilton.aspx

softworkz:
I know the article from Kim Hamilton and had even read it a few times in the past. If you read further into the topic, you will find, that many things in .NET do not work with the \\?\ notation. Would we start allowing this, lots of regressions would be guaranteed and the complete code would need to be reviewed. In certain situations, .NET code would need to be replaced with p/invoke, which is not portable, of course.

But it's not just the technical aspect. Would you really want that a multi-episode file gets a name like

_'Homeland S04E1-12 - The Drone Queen, Trylon and Perisphere, Shalwar Kameez, Iron in the Fire, About a Boy, From A to B and Back Again, Redux, Halfway to a Donut, There's Something Else Going On, 13 Hours in Islamabad, Krieg Nicht Lieb, Long Time Coming.mkv'_

I don't think this is a good idea anyway...

LukePulverenti:
Let me know thoughts on this as possibly a more simplistic alternative, that doesn't require the length param:

https://github.com/MediaBrowser/Emby/commit/cc5625c55991d3360bbab31fe372d36231f13704#diff-2ccebb4a8189fa1086827bf4a5ba2ae3R442

softworkz:
Your proposed solution is not reliable:
* It does not work when episode title is modified through _fileSystem.GetValidFilename
* It does not work with episode-name patterns %e.n and %e_n
* It may create files that start with a space character

softworkz:
Another suggestion (based on my commit): 

* Instead of adding parameter **int destPathLength** to GetEpisodeFileName, add parameter **int? maxLength** to this method
* Perform calculation of the maximum available filename length within the caller of GetEpisodeFileName

This way GetEpisodeFileName will remain portable (if that was the point of your concern), and the things I mentioned in my previous post would still be solved.

Finally one more thing: We need to subtract 4 more characters from the allowed length, otherwise subtitle downloading could fail... (Think of *filename.eng.srt*)

What do you think?

LukePulverenti:
i don't want to dwell on this for too long so just resubmit on dev and change my version and i'll merge in.

softworkz:
Hmpf, I screwed this, because I rebased against dev. I'll create a new pull request!

PR #1192, opened 2015-09-22
Auto-Organize: Added feature to remember/persist series matching in manual organization dialog

### LINKED ISSUE #1191: Auto-Organize: Series detection
Hi,

quite often I got situations where a series episode filename contains substrings that aren't eliminated out by the VideoParser's Clean method.

Example: "tvp-bones-S03E04.mkv"

After all processing, the series name is still "tvp-bones". As a consequence, the file cannot be auto-organized. And - even worse - all other incoming episode files will require manual interaction as well.
From my experience there are many other cases where the series name in the file does not match the series name in emby.

I have thought about two different solutions to this problem. The first one is easier to implement but might handle just a subset of the detection problems. The second one is more generic but involves a little bit more work:

1. deleted, was a bad idea

2. Fail initial detection like it is implemented at the moment. But: In the manual organize dialog add a  CheckBox option "In the future, organize all files containing 'tvp-bones' into the selected series"

softworkz:
When a filename cannot be auto-matched to an existing series name, the
organization must be performed manually.
Unfortunately not just once, but again and again for each episode coming
in.

This change proposes a simple but solid method to optionally persist the
matching condition from within the manual organization dialog.
This approach will make Emby "learn" how to organize files in the future
without user interaction,

To make this work, a new persisted property "AutoOrganizeNames" was
added to the BaseItem class, containing a list of series name strings
that should be considered as a series match.

In order to avoid creating a "stealth" feature which may lead to
unexpected match results with a growing match string list over time,
these match string are also exposed for editing in the
MetadataManager.(only for actual series entries; first item in region
"Metadata Settings".

I haven't done localization yet.

Please review - I am interested in your comments! At my site, it's
working fine and performing just like I have been wishing for quite a
long time...

PS: As I could see from the code, this feature has already been planned
some time. I just reused the existing variable "RememberCorrection" in
some places...

PPS: Resolves #1191


LukePulverenti:
This is a good idea but the data should not touch library items and metadata files. auto-orgnanize should be a self-contained feature and should keep all data to itself.

softworkz:
Why do you think it is bad to store these match strings in the same way as all oher data is stored?
Can you suggest an alternative way to store?
As far as I'm concerned, I do not care about where the match info is stored - but what I deem really important is a way to view and edit this information...

LukePulverenti:
in the auto-organize config file. imagine auto-orgnaize where a plugin and you have no control over the  core server.

softworkz:
But in this case, editing via the Metadata manager would not be possible, right?

LukePulverenti:
That's correct, and that's for the better. we don't want users having to jump around to different sections just to get a file organized. that's why the manual override is right there in the auto-organize ui. look at metabrowser for some inspiration.

softworkz:
Pardon, what's metabrowser? Is it different from Metadata Manager?

softworkz:
Hey Luke, 

please don't leave me alone with some vague suggestion here. It's your project, your rules, you are the architect!

Tell me how you want this feature to be implemented and I will do as requested...

LukePulverenti:
http://themetabrowser.com/

softworkz:
I still don't understand how metabrowser is related to auto-organize. Could you please post a screenshot of the UI you are refering to?

LukePulverenti:
i haven't used it in a few years so i don't have one. you can always install and check it out, but it had an equivalent, self-contained feature. you could add your own list of search and replaces to the file names to allow them to match up with series in your library. we don't need the ui for configuring that right now, just the remembering is fine. so if series name X is extracted from the filename, but the user overrides it to match series Y, then just have it remember those before and after names for use later.

softworkz:
Well, if the UI is of no interest, then there is no point in looking at metabrowser.
The workflow you described, is exactly what I had implemented anyway.
Now I have changed my implementation as you requested: The matching strings are now stored in the AutoOrganize config and the UI for deleting those strings is within the dashboard's auto-organize section.

The workflow is as follows:

1. A file fails Auto-Organization because the Item-(Series-)Name cannnot be resolved
2. The user can now organize the file manually from the Auto-Organize Log
3. I have added an option to this dialog to save the match string:

embysmartmatchorganize

4. If the option is checked, the match string is saved to the autoorganize.xml file and in the future, all files with this string are automatically matched to the library item

5. I have called this feature "Smart Match" and added a UI directly in the dashboard:

embysmartmatch

Currently match strings can be deleted only. Probably a method for adding custom strings, will not be required anyway. It is just important to be able to review the smart match configuration and to correct undesired settings.

What to you think?

LukePulverenti:
it's nice. just use generic text in the popup. it's easier on the translators. also, use the paper list style. check the notifications page in the latest code for an example.

softworkz:
Regarding translations: Can you suggest a tool to edit all string files in a synchronous way?
Regarding style: I am already using the existing styles  - actually I copied a few bits from the scheduled tasks page. What's do you think is wrong? 
PS: I just tweaked this a bit to get series names bigger and match strings smaller. But I will gladly leave the visual tweaking up to someone else..

LukePulverenti:
you just edit server.json for html files, javascript.json for js, and the language versions are generated by our translation site.

softworkz:
OK, good to know that. I tried some localization tool and it screwed up all files..

psxlover:
What if euhd was related to the series name and elefantastisch was an episode title in the first screenshot? Adding all of it as a smart match would make no sense since it would only match a single episode, it should probably be a text box.

Also in the second screenshot as can be seen by "Anger Management" using the whole string makes little sense, since it contains the release group, which can be different in each episode.

softworkz:
@psxlover: The important thing here is, that we are NOT matching against the filename. We are matching against the series name that emby has extracted through its regex patterns.

For this reason it is very unlikely that euhd is the series name and elefantastisch an episode name, because this is already the parsed result.

As to the second case: Yes, currently you would have to repeat the procedure for a title with a different release group. But cases with more than two or three different release groups are probably rare.

But I understand your idea and I have thought about that, too. The thing is: Before this change, emby was matching the regex-extracted series name by exact comparison: exractedName == seriesName.
My change stays with that method, it just adds alternate recognition strings.

What you are proposing would be a substring match. E.g.: the extracted name is "zzgtv-anger" and you would like the user to be able to change this through a checkbox to "anger". In order to make this work, the matching would need to be changed from exact matching to substring matching. This is a bit dangerous, because when there is another series containing 'anger' in the title, it would be incorrectly organized. You may say, that this is a rare case, but there are even more implications with substring matching: The watch folders may not just receive tv series episodes but also movies or songs or some other stuff. This makes it very likely that you may some time receive a file where the regex-extracted name contains the word "anger" - and this is why substring matching is a very bad idea in this case.

I thought about that for quite a while, but in the end I always came to the conclusion that substring matching would cause many more problems than it would solve. 
Of course, you will have to go through the organize correction dialog two times, if the files are from two release groups, but that is at least much better than doing it 24 times for a single season...

softworkz:
A future improvement I have in mind would be this: Once you have completed the correction dialog and checked the "Smart Match" option, the server could go through the failed log entries and retry organization. This would help in cases where you already have an amount of failed files in the log (e.g. all containing 'zzgtv-anger').

softworkz:
This feature is now also included in #1202

LukePulverenti:
Can we try to be more primitive and not have the smart match model classes? All of the settings for tv auto-organize should be self contained underneath TvFileOrganizationOptions. You should be able to reuse the NameValuePair class and just have a simple list of them.

softworkz:
I tried to implement this in an independent way where SmartMatch is not restricted to TV organization but rather a feature that could be shared by other organization methods which may be added in the future (at least in theory).
Since we are dealing here with a structure that is persisted to the users' configuration files, there is no easy way to change this in the future. I.e.: If we put SmartMatch data directly into TvFileOrganizationOptions instead of AutoOrganizeOptions, we would run into a problem later..

Regarding NameValuePair: I don't think that this is an appropriate structure for 1->n entities. Also, considering the resulting xml structure of the config file, I'd say that my approach produces better results...

I have employed some simplification of the storage just now - see the last commit...
But if you insist on NameValuePair - tell me and I'll do so!

LukePulverenti:
i don't envision this extending beyond tv so i want to simplify as much as possible. movies are always one off, case by case therefore it will not be as useful. maybe one day we'll organize music but even there it's not really as useful. i also don't see a one to many relationship here. just fill in the search string, then the replace string.

softworkz:
OK, I'll move it to TvOrganizeOptions.

Regarding one-to-many: Just look at the screenshot:

https://cloud.githubusercontent.com/assets/4985349/10037428/335a4236-61b6-11e5-875d-f3bd21a6e88f.jpg

softworkz:
Nonetheless, I think it would stiil be worth, talking about the auto-organize architecture:

With this change, I modified several function signatures to use AutoOrganizeOptions instead of TvOrganizeOptions.

The other thing I'm working on already does rely on this change:

* **EpisodeFileOrganizer** gets a new base class **FileOrganizerBase**
* A new class **MovieFileOrganizer** derives from this same class

As a consequence, the individual function signatures should not use a specific options class (like TvOrganizeOptions) but a generic options class instead (like AutoOrganizeOptions)

If you like I surely can move the smart match strings to TvOrganizeOptions. But what is the actual advantage over the alternate way? I squeezed my head but didn't get an idea...

softworkz:
This is an independent subset of changes included in pull request #1202 anyway. If #1202 would be merged, this PR could be closed.

LukePulverenti:
I'm ready to merge this one. Can you:

* sync to latest dev
* remove the de language text
* remove the javascript.json text

Thanks.

softworkz:
OK, hold on...

softworkz:
Hmpf, I need help. I rebased to dev and fixed up everything.
Then I did "git push --force"
Suddenly this branch is targeting "MediaBrowser:master". I'm sure it was targeting "MediaBrowser:dev" before.

I read in many places that retargeting a pull request is not possible and not supported, but in my case the target branch _HAS_ changed. That's kind of weird, at least..

softworkz:
This pull request (or rather its branch) seems to be seriously broken.
Thus, I created a new pull request #1432 to replace this one!

PR #1197, opened 2015-09-25
Auto-Organize: Display errors in client on server exceptions

softworkz:
When an exception occured on the server, the organize dialog or the
confirm dialog were displayed eternally and not indication of failure
was presented to the user


softworkz:
PS: An easy way to test this is to remove or rename a file that could not be auto-organized and then try manual correction

LukePulverenti:
Needs localization

softworkz:
How should this be done?
I can localize "Status:", but I cannot localize the errors that come from the server...

softworkz:
Please clarify: Do you mean just localization of the "Status" label? Or is there more to it?

ebr11:
Yes, just localization of the text you have control over.  See the other pages for how to do it.

softworkz:
OK, done!

LukePulverenti:
You can't do this

Error: {0} - {1}

because the translators will have no way of knowing how to handle that. You'll just have to use multiple lines.

softworkz:
It's the message box header, so no multiple lines.

LukePulverenti:
Then you can't do that. The header should only be 2-3 words anyway

softworkz:
Then I'll do
Header: "Error"
and move the rest to the message, OK?

LukePulverenti:
Maybe Error Organizing File would be a little friendlier.

softworkz:
OK, done!

LukePulverenti:
Maybe we will never support right to left but i have been avoiding text concatenations such as this just to give ourselves flexibility

e.status + ' - ' + e.statusText + '<br>' + e.getResponseHeader("X-Application-Error-Code")

Also what are some examples that come through from X-Application-Error-Code? That wasnt' really meant for user display.

softworkz:
If you want localized ordering of the elements (i.e. first *statusText* then *status*), then the previous variant using {0} and {1} (which you didn't like) would have been capable to achieve this.

Here are some examples where the X-Application-Error-Code header is needed to indicate the actual error to the user:

* User opens and completes the correction dialog for a file that does not exist anymore
* User opens and completes the correction dialog for an existing file but the file cannot be copied to the target location (drive or path not accessible, file in use, no access permissions, volume full, etc.)

In the current public release, the correction dialog just remains visible forever in all of these cases. This behaviour is even more irritating when you consider the fact that in case of a successful completion, you will experience a similar result: The correction dialog remains visible until the file move operation is completed. As a consequence, the user keeps waiting and waiting assuming that he's waiting for the file move operation to complete while in reality the operation has failed long ago.

I think it is important in these cases to indicate such failures to the user not only as a generic message (like "An error has occured") but also provide some useful information about this error. The X-Application-Error-Code provides exactly this, so I did not look any further.

softworkz:
Retargeting a pull request seems to be a no-go with GIT.
Will create a new pull request.
Closed.

PR #1199, opened 2015-09-25
Auto-Organize: Perform leftover-file and empty-folder deletion in processed folders only (#2)

softworkz:

### DIFF
--- a/MediaBrowser.Server.Implementations/FileOrganization/TvFolderOrganizer.cs
+++ b/MediaBrowser.Server.Implementations/FileOrganization/TvFolderOrganizer.cs
@@ -98,35 +98,14 @@ namespace MediaBrowser.Server.Implementations.FileOrganization
 
                 if (options.DeleteEmptyFolders)
                 {
-                    foreach (var subfolder in GetDirectories(path).ToList())
+                    if (!IsWatchFolder(path, watchLocations))
                     {
-                        DeleteEmptyFolders(subfolder);
+                        DeleteEmptyFolders(path);
                     }
                 }
             }
 
             progress.Report(100);
-        }
-
-        /// <summary>
-        /// Gets the directories.
-        /// </summary>
-        /// <param name="path">The path.</param>
-        /// <returns>IEnumerable{System.String}.</returns>
-        private IEnumerable<string> GetDirectories(string path)
-        {
-            try
-            {
-                return Directory
-                    .EnumerateDirectories(path, "*", SearchOption.TopDirectoryOnly)
-                    .ToList();
-            }
-            catch (IOException ex)
-            {
-                _logger.ErrorException("Error getting files from {0}", ex, path);
-
-                return new List<string>();
-            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


LukePulverenti:
Please resubmit on the dev branch. Thanks.

softworkz:
I really tried and tried again, but in the end, it always said "Can’t automatically merge". I really don't know why, because there is no actual conflict in the file. 
Please help me, before I start hating GIT ;-)

psxlover:
[TvFolderOrganizer.cs](https://github.com/MediaBrowser/Emby/blob/dev/MediaBrowser.Server.Implementations/FileOrganization/TvFolderOrganizer.cs) was modified 3 days ago in the dev branch, so github does not know how to merge your version of TvFolderOrganizer and dev branches version. What you need to do is rebase your branch on the remote dev branch. I also recommend that you remove the merge commit from your commit log before doing the rebase cause it may cause you issues during the rebase.

softworkz:
@psxlover Thanks for your help! The problem is that this change is in my forked master branch and I have several feature branches based on this master branch. When I rebase my master branch, will this affect my other feature branches? Or is it save to do it?

psxlover:
Any changes you make in a branch (including rebasing) wont affect any other branch so it's safe to rebase. If you are afraid you might break something, create a new branch from your master and rebase that.
Note that your "origin" probably refers to your fork, and not to this repo. You'll have to add a new remote pointing to MediaBrowser/Emby instead of softworkz/Emby, and rebase on that remote's dev branch.
In git even if you do break a branch (for example if you delete a branch, or you do a hard reset) the old commits are still there for some time (I think the garbage collector will delete them only after a couple of months has past) even if you don't see them anywhere. As long as you know the old commit hash you can create a branch using that hash and the log will be exactly as it where. You can find md5 hashes of "deleted" commits using the reflog.

softworkz:
Thanks again. I already had found out how to add the original Emby as upstream project in order to rebase, but I wasn't sure about the consequences on the other branches...
One last thing: When I create a new branch including this change and rebase that branch, can I update this pull request to use the changes from the new branch, or is the only way to close this pull request an create a new one?

softworkz:
And finally: In the future, should I always submit pull requests against the dev branch, or just this one?

psxlover:
Unfortunately you can't change the branches of a pull request in Github (I believe there is a feature request about this but hasn't been fulfilled) , so you either had to replace your master branch and reopen PR #1200 which was targeting "dev" branch or create a new PR.

>And finally: In the future, should I always submit pull requests against the dev branch, or just this one?

That depends. Unless the project changes its release cylce (this is probably not the correct term but I can't think of the correct one at this hour) you should submit your PRs against the dev branch and not the master branch. Looking at the log I believe that the dev branch is used for anything new, and everything is merged in the master branch only when a new stable or beta version is released (I could be wrong about this though).

softworkz:
I suspected this, that you can't change a pull request's source branch, so I took the way through a new pull request. The inconvenient thing with this procedure is, that all discussion that the issue may have gone through lives in a different place then (like in this closed PR).

Thanks for your thoughts about the targeting branch. I was unsure since all my pull requests were against master and nobody complained...

@LukePulverenti So one should always target pull requests to the dev branch, right?

ebr11:
Yes, you should always be working in dev. Thanks.

softworkz:
OK, will do!

PR #1200, opened 2015-09-27
Auto-Organize: Perform leftover-file and empty-folder deletion in processed folders only (#3)

softworkz:
 DeleteLeftOverFiles and DeleteEmptyFolders functions


softworkz:
Resubmit of #1199 against dev branch.

PR #1201, opened 2015-09-28
Auto-Organize: Perform leftover-file and empty-folder deletion in processed folders only (targeting dev branch)

softworkz:


softworkz:
This is a follow-up to pull request #1199, targeting MediaBrowser:dev instead

softworkz:
Thanks! :-)

PR #1202, opened 2015-09-28
Organize Movies

softworkz:
**Working Prototype!!!**

This branch adds multiple new features to Emby's Auto-Organize capabilities:

1. Added an option to remember auto-organize corrections: This is further described in pull request #1192. In addition to the #1192 code, the option is only displayed if a non-empty series name was parsed

2. Enables manual organization of movies like described in this thread or also here: http://emby.media/community/index.php?/topic/25857-movie-organization/  
This is successfully implemented now, but I wouldn't claim this to be anything better than just a working prototype!

3. Enable addition of a new series from within the auto-organize form: You can search for a series, view search results and select an item. The current episode is then organized as a new series. You can select a root folder for the new series from a dropdown containing all of your library TV folders.

This branch also includes the changes from #1209. 

In the context of #1209 I did some further improvements: When an AJAX request takes longer than 30s, the request fails with status==timeout. 
Before #1209 the failure was not indicated to the user. Instead, the loading spinner graphic (and the dialog, if initiated from it) was shown forever.
After #1209 any AJAX failure was presented to the user via Dashboard.alert and so did the timeout message.
But when organizing a video file manually, which may have a size of several GBs, the process often takes much longer than 30s.
To improve the user experience in these cases, I 

* added a specific message which is displayed on timeout errors ("The operation is going to take a little longer. The view will be updated on completion.")
* added an additional websocket message to inform all clients about updated auto-organize data. Previously the client updated the auto-organize log only when an auto-organize task had finished completion. Now, the client also updates the auto-organize list when an organize-function on the server has completed.


softworkz:
I am thinking about extending the Auto-Organize-Correction dialog in order to be able to handle movies.

While I have doubts if an automatic organization of movie files would even be possible in a useful way, we could at least offer an easy method to do it manually. My current prototype looks like this:

automovie

What do you think about this? For myself it would be a very useful feature. No more manual copying files and creating folders, and no more need to perform a manual identification in case it doesn't work out automatically...

softworkz:
This is how the "Create Series" feature is supposed to work:

movieorganization3

* Click on "New Series"
* You will get a dialog similar to the one for movie identification:

movieorganization4

* Enter/correct series name
* Click "Identify"
* Choose series from search results or click "Cancel"

Now you would click "Create Series" and be returned to the initial dialog with the newly created series selected in the series dropdown field.
**Beware: This last step is not implemented yet!!**

softworkz:
Rebased to and adapted for dev/HEAD

softworkz:
Completed the "New Series" feature for the auto-organize-correction dialog.

LukePulverenti:
is that available separately without movies?

softworkz:
No. These two features are sharing quite a few things...

softworkz:
But you could just hide the tab switch element:

autoorganizetabswitch

LukePulverenti:
can you change it to a dropdown instead of the tabs? also i saw some hardcoded english text somewhere in the javascript.

softworkz:
PS: I spent endless hours trying to do this with paper-tabs. In the main branch this worked without problems, but not in the dev branch. Paper-tabs are working just fine in the normal html, but not in a popup. I even tried to migrate the jQuery popup to Paper-Dialog, but same result: Paper-tabs didn't work...

softworkz:
Yes, localization is still pending!

LukePulverenti:
the reason i want the tabs is to keep user decision making strictly in the vertical direction. Also the identify button in the middle is kind of clunky. It's unclear if i need to click it or not. If it's required, then every field that comes after should be unavailable until the identify process is completed. That way there's only one button to click on. If it's not required, then we need to decrease it's level of prominence  and maybe just turn it into a search icon button next to the name field.

Can you also change the style of popup to what is being used by the normal identify popup? if you do that, then you'll have no problem switching to the newer paper elements, and there will be a search icon button that can be used.

softworkz:
> the reason i want the tabs is to keep user decision making strictly in the vertical direction. 
> Also the identify button in the middle is kind of clunky. It's unclear if i need to click it or not.
> If it's not required, then we need to decrease it's level of prominence and maybe just turn it 
> into a search icon button next to the name field.

I abolutely agree! I think the "New Series" button should be a small button right to the series dropdown. I just didn't do it because I couldn't find a precedence case in the existing UI markup which I could adapt accordingly. Multi-targeting html code is out of my focus anyway, so I wasn't sure if horizontal stacking is allowed at all..

> If it's required, then every field that comes after should be unavailable until the identify process is completed. That way there's only one button to click on. 

I agree. But I couldn't find an appropriate "disabled" style for the input.

> Can you also change the style of popup to what is being used by the normal identify popup? 
> if you do that, then you'll have no problem switching to the newer paper elements, 
> and there will be a search icon button that can be used.

The paper-dialog was just a local experiment. In the submitted code there is no change to the popup method. I also don't have any individual popups. It's always the correction dialog with variable content...

LukePulverenti:
You just use plain html disabled attributes. But I think it would be better for the fields not to be visible, problem is the jquery mobile popups don't react well to changing height.

I know you're just preserving the popup as it was, but I think this feature really needs to switch to the newer methods because i've been in process of deprecating a lot of jquery mobile features like have a nice small icon button or put icons on the text buttons. also the newer style popup puts a back button in the top left which will allow the cancel button to be removed. It's really not difficult. here's an example:

https://github.com/MediaBrowser/Emby/tree/dev/MediaBrowser.WebDashboard/dashboard-ui/components/itemidentifier

and here's how it's called

https://github.com/MediaBrowser/Emby/blob/dev/MediaBrowser.WebDashboard/dashboard-ui/scripts/librarybrowser.js#L947

It will wind up being very large but that's ok. i can handle that later.

softworkz:
I had a look and found that I can't use the itemidentifier as a component. It is built specifically for the library identification process (displaying the identify options and calling "Items/RemoteSearch/Apply/").
Also dealing with multiple stacked popups -- I don't know how well that would work. That whole html/css/jss stuff is among the top 3 on my hate list :-)  - So I wanted to avoid changes that may break something.

But during my frustrating efforts to get paper-tabs working, I already had migrated the popup to paper-dialog once - so I could at least do that again and also replace the cancel button, if that's OK?

LukePulverenti:
Right you can't use it i"m just saying use it as inspiration to create a new one.

softworkz:
OK, understood! Then I'll leave it all in a single popup as paper-dialog. On Chrome is resizes fine, but if it doesn't on other platforms, there are still appropriate methods provided by the paper-dialog to fix this. Right?

LukePulverenti:
well no, not leave as paper dialog, look at the source of the identify "component" and design this one in the exact same way. if you do that then the sizing point is moot because it has a fixed size and scrolls when needed. you might even prefer instead to look at components/imageuploader because that is the exact same thing and only 190 lines of code. so that will be quicker to copy paste and remove what you don't need.

softworkz:
So you are saying I should implement the complete auto-organize popup as a component?

LukePulverenti:
you don't necessarily have to. if you look at imageuploader.showEditor, that is basically the code needed to display the popup. you can then put the popup markup into a template.html file just like it's doing. so you don't actually need to move this into the components folder.

softworkz:
So, the important thing is that the popup markup is not contained in the actual page html?
Probably a good idea since the jQuery popup is sometimes visible for a short time when the page loads. And I can get rid of the code to reset all form fields for subsequent popup displays.

But in the normal html folder there are no other .template.html files... Wouldn't it be more consistent, to put it in a component subfolder, then? (or in a <template> tag)

softworkz:
Sorry for my detailed asking. I just don't wanna do it twice or thrice...

* Should I also migrate all form inputs to the corresponding paper-versions?
* How can I arrange the "New Series" button horizontally to the right of the series dropdown? (I mean in a portable way... no problem to make it work in a desktop browser).

LukePulverenti:
well the main important thing is the new popup style. whether it goes into the components folder or some quarantined auto-organize folder doesn't really matter so much. yes migrate to the paper inputs.  you can make the series dropdown width 80% and use paper icon button. 

fyi, the popup will be huge and will have a dark background but that's ok, stick to the pattern. i will adjust after merging it in.

softworkz:
**The good, the bad and the ugly:**

I followed the component pattern. Works fine so far! 

* The good: paper-tabs are working again with this pattern! I migrated back to paper-tabs!
* The bad: While Polymer may be a newer and better approach for html multi-targeting, it has gotten even harder once again to investigate, debug and identify what's really going on when things are not working as expected
* The ugly: Width: 80%? This doesn't even succeed with a single browser - as long as you don't write some dedicated code to adjust this. Or am I missing something here? The requirement would be as follows: 1. Button width should be constrained to a fixed minimum determined by the caption text. 2. Dropdown width should expand to fill the remaining space. On all supported platforms... I doubt if that could really be achieved without writing js code....

My next step will be to replace the html inputs with Polymer controls....

softworkz:
Migrated most form inputs to paper-controls. 
Performed new vulcanization because paper-dropdown-menu was missing.
Replaced Cancel buttons in favor of the new back button.

LukePulverenti:
well skip paper dropdown because we're not using that anywhere in the application. sorry, should have been more clear about that. i don't like their implementation of that at the moment so we're not using it yet.

softworkz:
Hmpf, spent quite a but of time on that...

softworkz:
Are you really sure that Polymer is ready for productive use?
I encountered so many problems - just using Chrome, which should be the best for Polymer...
I wonder how that works on other browsers (desktop and mobile)...

LukePulverenti:
we're only using a limited subset that i've specifically tested.

softworkz:
But when mixing paper-input controls with classic controls (like SELECT), there will be a different look&feel...
This is the latest screen shot:

organizemovie9

The paper-inputs have that special super-duper-feature to display their label in a big version within the input while the input is empty and this label shrinks and moves to the top of the input once the user has entered some data. According to the docs this behaviour can be deactivated by adding an attribute "alwaysFloatLabel" which does not actually work, though. But it works, when you add an "always-float-label" attribute.
Having to check out such things is burning huge amounts of work time. I'm not really sure if Polymer is such a good alternative (at least at this time).
When you are saying now, that html option select should not be migrated to its corresponding paper control, we will end up with mixed controls of different look&feel. Then you would have to invest even more time to make these look and behave at least in a similar way. Is it really worth the effort?

LukePulverenti:
it is a gradual transition. if you just stick to the patterns currently used you would not had to spend time on any of that stuff. we are not using the label float attribute anywhere. I apologize, I thought this would be largely cut and paste from the identify popup without any research required. If you want to just submit what you have i will take it from here and complete it.

softworkz:
Yes, please, I would be really grateful if you could do that!

Organize Movies #8 was the transition to paper-dialog and
Organize Movies #9 was the migration to paper-input (and switching to the back-button)

So you could use everything until and including "Organize Movies #8" and use just those parts of "Organize Movies #9" that you would consider useful...

softworkz:
The previous attempt (Organize Movies #9) to migrate all controls to paper-controls didn't work out well, so I reverted that change.

This commit should get it all working again!

The switch to paper-dialog as a component (technically) and the new dialog layout style (back button in the upper corner replacing individual cancel buttons)  is completed now and most of the obvious bugs should be fixed.

softworkz:
Improve AutoOrganize-Feedback 

This commit improves client feedback for auto-organization.
As a result, the organize-log is updated for each item separately, not just as a single update once the organize task has completed.
This change depends on the introduction of the new websocket message "AutoOrganizeUpdate" inside this branch, so I can't separate it into a different pull request...

This change also fixes the Organize-Taskbutton functionality in the auto-organize log (part 1: adjusted style of progress bar; part 2: added basic execution delay)

softworkz:
Update auto-organize log in all clients on single log item deletion or clear log action

When two or more clients are displaying the auto-organize log at the same time:
* When deleting an item, all clients  are instantly updated now
* When clearing the log, all clients are instantly updated now

(This change depends on the introduction of the new websocket message "AutoOrganizeUpdate" inside this branch, so I can't separate it into a different pull request...)

softworkz:
...rebased to current dev branch.

softworkz:
Added server side locking to prevent duplicate item processing and changed the autoorganize list to display a processing indicator.

Please see the following video: https://youtu.be/Lq4RGXVh_os

softworkz:
AutoOrganize: Instant completion of correction dialog 

When the autoorganize correction dialog is submitted, the server service waits just 2s to complete the request.
These 2s are to wait for an exception that might occur. In this case the exception is forwarded to the client and the web UI will immediately display a message overlay with the error message.
Otherwise the ajax call is returned after 2s and the dialog gets closed in the client UI. The organization progress is then displayed in the autoorganize list (orange text color and spinner animation).

softworkz:
Once again, rebased to latest dev...

softworkz:
AutoOrganize: Make LeftoverDeletion and EmptyFolderDeletion also work with manual organization:

Previously the auto-organize features *LeftoverDeletion* and *EmptyFolderDeletion* only worked when organization was performed through the internal scheduled task.
This commit makes these features also work for all manual organize operations:
* Organize via correction dialog (series or movie)
* Retry organization when target is already set (series or movie)
* Delete file

To make this work in various situations, the code for these two features was refactored into *FileOrganizerBase.cs* without functional changes.
This commit also improves the automatic search term for movie organization.

LukePulverenti:
Hey,
This is great stuff, although can anything like the instant feature be extracted to a separate PR?

softworkz:
Over the time I always tried to extract certain changes to a separate branch.
Sometimes it works, but sometimes it doesn't. I always tried to isolate changes as good as I could, but in this branch there are too many changes where many of them are depending on others. 
I am testing my code every day on quite a few real world scenarios. I think I have achieved a state which is rather stable and really worth looking into it.
I know that this has grown quite a bit over the time, but looking at the details, it is not that much stuff that has changed.

To help you a bit looking through the changes, I have compiled a minimal list explaining all of my changes. It should be in the same order as it is presented here on github.

softworkz:
•	MediaBrowser.Api/Library/FileOrganizationService.cs
o	Added properties for new series organization
o	Added method for movie organization
o	Added methods for Smart Match entry management
•	MediaBrowser.Controller/FileOrganization/IFileOrganizationService.cs
o	Added Dictionary for InProgressItems
o	Added methods for movie organization and for Smart Match entry management
•	MediaBrowser.Model/FileOrganization/AutoOrganizeOptions.cs
o	Added SmartMatch options via composition
•	MediaBrowser.Model/FileOrganization/BaseFileOrganizationRequest.cs
o	Created as a base class for the existing EpisodeFileOrganizationRequest and the new MovieFileOrganizationRequest
•	MediaBrowser.Model/FileOrganization/EpisodeFileOrganizationRequest.cs
o	Added properties for new series creation feature
•	MediaBrowser.Model/FileOrganization/FileOrganizationResult.cs
o	Added some new properties: IsInProgress for dynamic progress display in the client and two others for movie organization
•	MediaBrowser.Model/FileOrganization/MovieFileOrganizationRequest.cs
o	New class for movie organization
•	MediaBrowser.Model/FileOrganization/SmartMatchInfo.cs/ SmartMatchOptions.cs
o	New configuration classes for the SmartMatch feature
•	MediaBrowser.Server.Implementations/FileOrganization/EpisodeFileOrganizer.cs
o	Made subclass of the new FileOrganizerBase
o	Changed  options usage up one level (AutoOrganizeOptions instead of TvFileOrganizationOptions) to enable access to SmartMatch options
o	Parse and store movie naming to optionally allow treating a file as a movie
o	Already save result at the beginning of the function to enable dynamic display of organization progress
o	Also perform LeftOverDeletion and EmptyFolderDeletion on manual organization operations
o	Implemented SmartMatch feature to remember certain strings as item identification
o	Improved error handling and error display – generally and especially in PerformFileSorting()
o	Implemented series matching via SmartMatch strings
o	Fixed bug when providers don’t return episode titles
•	MediaBrowser.Server.Implementations/FileOrganization/FileOrganizationService.cs
o	Added ConcurrentDictionary for items that are currently being processed
o	All returned FileOrganizationResult objects are updated with a valid IsInProgress property
o	Can now handle episode as well as movie corrections (depending on result.Type)
o	Implemented a new Websocket message "AutoOrganizeUpdate", to tell the client to update its organize list
•	MediaBrowser.Server.Implementations/FileOrganization/FileOrganizerBase.cs
o	Base class for the existing EpisodeFileOrganizer and the new MovieFileOrganizer
o	All methods  which are common to both variants were moved here
o	Implemented cleanup of parsed movie names for movie organization
o	Implementation of features “LeftOverDeletion” and “EmptyFolderDeletion” was moved here
•	MediaBrowser.Server.Implementations/FileOrganization/ItemInProgressException.cs
o	New custom exception which allows specific handling of this exception in certain situations (=> throw error on manual organization, ignore errors and continue during auto-organization as task)
•	MediaBrowser.Server.Implementations/FileOrganization/ItemProgressLock.cs
o	New class to implement locking and status of items currently being processed
o	Implements IDisposable to enable usage with minimal code by employing the using-pattern
o	Throws ItemInProgressException is item is locked already
•	MediaBrowser.Server.Implementations/FileOrganization/MovieFileOrganizer.cs
o	New class implementing (manual) movie organization
o	Currently the primary file is copied/moved only. This should be improved a bit in the future
•	MediaBrowser.Server.Implementations/FileOrganization/OrganizerScheduledTask.cs
o	Just added some new dependency services
•	MediaBrowser.Server.Implementations/FileOrganization/TvFolderOrganizer.cs
o	Added some new dependency services
o	Removed “LeftOverDeletion” and “EmptyFolderDeletion” features which were moved to FileOrganizerBase
•	MediaBrowser.Server.Implementations/IO/FileRefreshItem.cs
o	Item class for the new refresh queue
•	MediaBrowser.Server.Implementations/IO/FileRefreshQueue.cs
o	Timed queue class refactored out from LibraryMonitor, now grouping all items by parent folder and independently processing these groups
o	This approach is much more reliable/stable
•	MediaBrowser.Server.Implementations/IO/LibraryMonitor.cs
o	Separated queue stuff into FileRefreshQueue
o	Remaining code is very simple now
o	FileRefreshQueue could be used for watching autoorganize folders (instead of polling)
•	MediaBrowser.Server.Implementations/Persistence/SqliteFileOrganizationRepository.cs
o	I added two additional fields to the auto organize table to store the parsed movie name and year just in case that it is a movie that we are dealing with
o	I did this for the following reasons:
	The SqLite variant provides a proven storage method from which I didn’t want to deviate, replace or amend
I just didn’t want to come up with something where people could have thought that I wanted to outnumber anyone else or overtake the project lead
	Code changes were minimal for this, so it jumped right into my head
o	Alternative: Generate these strings just on the fly and don’t store them in the db table
•	MediaBrowser.Server.Startup.Common/ApplicationHost.cs
•	Minimal changes
•	MediaBrowser.WebDashboard/dashboard-ui/apiclient/apiclient.js
o	Added methods for SmartMatch and movie organization
•	MediaBrowser.WebDashboard/dashboard-ui/autoorganizelog.html
o	Refactored correction popup into component
•	MediaBrowser.WebDashboard/dashboard-ui/autoorganizesmart.html
o	Management of new SmartMatch feature
•	MediaBrowser.WebDashboard/dashboard-ui/components/fileorganizer/fileorganizer.js
MediaBrowser.WebDashboard/dashboard-ui/components/fileorganizer/fileorganizer.template.html
o	UI for new organize-features
•	MediaBrowser.WebDashboard/dashboard-ui/scripts/autoorganizelog.js
o	Removed correction functions
•	MediaBrowser.WebDashboard/dashboard-ui/scripts/autoorganizesmart.js
o	Functions for new smart match feature

softworkz:
Sorry, that looks aweful. I'll try to find somethin better..

softworkz:
OK, I think I got it now - should be readable...

softworkz:
Rebased to the current dev branch...

danergo:
softworkz, I'm really interested in your work here.
How can I use it with the master branch? (dev branch has issues on Mono, it's crashing right after start due to some network interface implementation failure).

What do you think, is it possible to merge your changes into master via patch?

Thank you.

softworkz:
Danergo,

this would be a complicated task. It's not just about the 44 files where I made changes. There are many more changes in the dev branch which my changes are relying on (just thinking of th Polymer components for html rendering).

Sooner or later these changes will be merged from dev to main, but in a consistent order, and then my changes will fit into the main branch as well. 

I we would create a branch from main and try to merge only my changes into it, we would need to start cherry-picking those changes from dev that my changes are relying on. But there are chains of dependencies that would need to be resolved manually with great effort, to pick all required changes. In the end, you might find that almost all dev branch commits must be included and the issues with Mono are inherited as weil...

So not, it's not (economically) possible this way!

The way to go is to fix the Mono issue in the dev branch. What problem do you have? Did you create an issue?

Regards,

softworkz

danergo:
I can understand you completely, thank you for the comprehensive feedback.

My only problem is that I can build the dev branch, however upon running I
get an error on startup for using Socket.set_DualMode.

Do you know how can I get through on this error?

Thank you

softworkz:
I've seen your conversation with Luke at Emby Community. He is the right guy for this problem!

danergo:
Hi, softworkz!

Do you have any news on the merge? I'm seeing green lights everywhere, but still have no idea weather or not it has been merged. :)

Thanks

softworkz:
Unfortunately it hasn't been merged yet.
Luke is very busy at the moment and is plannng to merge my changes manually.
I hope he will find some time soon.
In the meantime I am trying to keep this up-to-date with all other changes in this branch.
At my site it is running great so far. I successfullly organized many new series and movies yet...

danergo:
Thanks for the feedback.
I found another problem around auto-organizing.
This is for series. I created the main series folder and also the season folder inside.
However, auto-organize couldn't resolve that, because I created "The.Big.Bang.Theory", and the files were named like "the.big.bang.theory". Will your modification fix also this?
Because it will try to recognize the series name from the files, right?

Thanks.

softworkz:
* I didn't have this series in my library
* I created a file "The.Big.Bang.Theory S01E01.mkv"
* In the Auto-Organize-Dialog the series name was detected as "The.Big.Bang.Theory" 
* I chose "Create new Series" and clicked on "Identify" (without removing the dots)
* The series was found and I selected the entry 
* The folder name was set to "The Big Bang Theory (2007)"
* The folder was created and a "Season 1" folder was created inside
* The episode was moved to this folder and renamed to "The Big Bang Theory - S01E01 - Pilot.mkv"

**I'd say, so far it worked just as it should!**

* After another library scan task had completed, the new series was recognized correctly and appropriate images were downloaded for series and season
* Then I added another file "The.Big.Bang.Theory S01E02.mkv" to the auto-organize folder
* It was matched to the series automatically (despite the dots), so in this case the smart match feature (which is new as well) wasn't even required

(SmartMatch would help in cases where files are named like "TBBT.S01E01.mkv")
**This behaviour appears to be fine as well!**

* I also tried creating a series folder named "The.Big.Bang.Theory" manually and added files like "the.big.bang.theory...".
* In this case the episode was correctly identified as well

**Thus, I suppose this is either fixed by "my" changes or - if not - we both tested this under different conditions..**

danergo:
@softworkz:
Do you know when will this be merged into the dev branch? Or maybe @LukePulverenti ?

Anyway, thank you for the previous feedback, I just didn't get any notifications for somewhy.

Keep up, guys!
danergo

softworkz:
Smart Matching is in the current beta and dev.
New series creation is in the current dev.
Movie organization will hopefully follow soon..

danergo:
Thanks, that sounds great :)

Very nice work, much appreciated!

danergo:
Hi, guys!

What do you think, when could this MovieOrganization request be merged into dev branch?
Can I help you anyhow?

Thanks,
danergo

softworkz:
I'm waiting for a number of fixes to be merged before I can go on with the movie stuff.
It's up to Luke when he'll do that

LukePulverenti:
I'm closing this. If we're going to do it, it will have to be as a plugin.

softworkz:
This feature has been implemented in a plugin:

http://emby.media/community/index.php?/topic/35246-new-plugin-organize-movies-with-movieorganizer/

PR #1204, opened 2015-10-01
OmdbItemProvider: Use search instead of title API

softworkz:
In my configuration, the Omdb provider is the first in the list for
movie metadata. This was the default at the time I installed MB and I
never changed that (don't know what the current defaults are for a new
installation).
When I use the identify command for movies in the metadata editor, I
always get a single result only. This is due to the fact that the Omdb
provider used the title API with "t=moviename", while there is also a
search api with "s=moviename" which will return multiple results.

This commit modifies the OmdbItemProvider to use the search API variant
and enable returning multiple results

### DIFF
--- a/MediaBrowser.Providers/Omdb/OmdbItemProvider.cs
+++ b/MediaBrowser.Providers/Omdb/OmdbItemProvider.cs
@@ -59,6 +59,8 @@ namespace MediaBrowser.Providers.Omdb
 
         public async Task<IEnumerable<RemoteSearchResult>> GetSearchResults(ItemLookupInfo searchInfo, string type, CancellationToken cancellationToken)
         {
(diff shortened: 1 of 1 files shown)


softworkz:
I should note that there is a slight drawback: The search API does not return images. But when trying to identify a movie, a single result is not helpful, even if it comes with an image...

LukePulverenti:
can you submit a request to the developer to add it? thanks.

softworkz:
OK, I will try! If he'll do, we are still set - no additional change would be required..

softworkz:
OK, I sent a message to the owner of omdbapi. I sent you a copy as PM on http://emby.media/community (had no other idea how to send you something)

Let's hope he will agree - at least I had a good argument...

PR #1205, opened 2015-10-01
Remote-Search: Allow result aggregation of multiple providers

softworkz:
Previously, when a remote search (without provider restriction) was
executed, the search used results from the first provider that returned
at least a single result only. Other providers were ignored.

This commit changes the behaviour in a way that all available providers
are queried until a certain number of search results has been collected.
The number is hardcoded to 10 (like it was before), but could be
parametrized in the future.

### DIFF
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -727,6 +727,8 @@ namespace MediaBrowser.Providers.Manager
             where TItemType : BaseItem, new()
             where TLookupType : ItemLookupInfo
         {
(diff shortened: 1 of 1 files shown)


LukePulverenti:
we need to make a revision to this because now the results are showing duplicates. when you add the results to the list, check the ProviderIds property and if there is a single existing provider id that matches, then skipping adding that item to the result list. This should work for Omdb + Tmdb and Omdb + Tvdb because they both have imdb in common.

softworkz:
OK, will do!

softworkz:
I'll also aggregate the image links. Sometimes one result does have an image while the other doesn't..

PR #1209, opened 2015-10-02
Auto-Organize: Display errors in client on server exceptions

softworkz:
This is a resubmit of pull request #1197 against the dev branch:

When an exception occured on the server, the organize dialog or the confirm dialog were displayed eternally and not indication of failure was presented to the user


softworkz:
This change is now also included in #1202

PR #1210, opened 2015-10-02
Auto-Organize: PathTooLongException on source file should not break auto-organize task

softworkz:
PathTooLongException can not only occur with long destination paths but
also with too long file names of files contained in a watch folder.
Previously this condition caused the auto-organize task to break.

With this change, we still log the exception, but auto-organize
processing will continue to handle all other files.

### DIFF
--- a/MediaBrowser.Server.Implementations/FileOrganization/TvFolderOrganizer.cs
+++ b/MediaBrowser.Server.Implementations/FileOrganization/TvFolderOrganizer.cs
@@ -35,7 +35,7 @@ namespace MediaBrowser.Server.Implementations.FileOrganization
             _providerManager = providerManager;
         }
 
-        private bool IsValidVideoFile(FileInfo fileInfo)
+        private bool FilterValidVideoFile(FileInfo fileInfo)
         {
             try
             {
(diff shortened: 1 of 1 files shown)


softworkz:
Please see also: http://emby.media/community/index.php?/topic/25946-auto-organize-has-stopped-working/

LukePulverenti:
How about instead give the function a name to indicate it's a filter to determine what files to organize. then move the length check in there as well. the purpose of the try/catch will then be to handle any possible error when querying the file info and returning false if something is encountered.

softworkz:
Oh well, the length check is already performed in the function *IsValidVideoFile*:
In case of a too long path, the FileInfo.FullName property will cause the exception.

If you are referring to the minimum file length check - these two are actually very different issues:

* Path too long: Throw and log error but continue processing all other watched files
* File too small: Silently ignore since it may be just a trailer or sample file

softworkz:
But of course I could rename it to *FilterValidVideoFile*...

softworkz:
I renamed the function to FilterValidVideoFile as requested.

softworkz:
Thank you! :-)

PR #1211, opened 2015-10-05
Added "+" button to TV Show page on Web Dashboard

mtlott:
- Allows creation of new TV Series from Web Dashboard instead of having to manually add folders through the native OS
- Fully refreshes meta-data and redirects to new series in web ui
- Added CreateSeries to TVSeriesManager
- Added /Library/Series POST to LibraryService
- Added tvshoweditor.js for side panel implementation
- Code is Modeled on Collection workflow


LukePulverenti:
Thanks. This is certainly very interesting but this upcoming release is all about simplification and removal of multiple ways of accomplishing the same thing. So don't take it personally but I may sit on this for a little while to see if this is something we really want to do. I love the contribution though and if you want to help elsewhere i can certainly give you some ideas.

mtlott:
Take what you like, leave the rest. As I add things that I need I will toss them over the fence. If you like it, its all yours. I am curious. How do you create a series from the user interface? I couldn't find anything that does this. I have to remote desktop back to the server to create folders at present. What did I miss?

PR #1219, opened 2015-10-15
Reworked library refresh implementation

softworkz:
The LibraryMonitor was implemented in a way where a single locked file
path queued for refresh kept blocking all other queued paths from being
refreshed.

An example why this was really bad: I have configured my WMC "Recorded
TV" folder as a library folder. Regardless of the time of day - in most
cases there is at least a single tv recording active. The LibraryMonitor
detected the new recording and added it to an internal queue for
processing. But processing does not start as long as even just a single
file in the queue is still locked (=being written).

After this change, all non-locked files are identified and processed
immediately while only locked files are deferred for later processing.

### DIFF
--- a/MediaBrowser.Server.Implementations/IO/FileRefreshEventArgs.cs
+++ b/MediaBrowser.Server.Implementations/IO/FileRefreshEventArgs.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MediaBrowser.Server.Implementations.IO
+{
+    class FileRefreshEventArgs : EventArgs
+    {
+        private FileRefreshItem _item;
+
+        public bool Cancel { get; set; }
+
+        public FileRefreshItem Item { get { return _item; }}
+
+        public FileRefreshEventArgs(FileRefreshItem item)
+        {
+            _item = item;
+        }
+    }
+}
--- a/MediaBrowser.Server.Implementations/IO/FileRefreshItem.cs
+++ b/MediaBrowser.Server.Implementations/IO/FileRefreshItem.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MediaBrowser.Server.Implementations.IO
+{
+    class FileRefreshItem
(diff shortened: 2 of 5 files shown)


ebr11:
Recorded TV probably shouldn't be defined as a library but, in relation to this specific change, the issue will be with the standard movie structure of each one in its own folder and many different files within that folder.  We don't want a refresh firing until all the files in that folder have finished being modified.

Did you test this scenario?

softworkz:
I understand. I'll revise it to group by folders...

softworkz:
But on the other hand: When grouping by folders: in case of a series folder, the library update would again be suppressed, when episodes are coming in continuously.
Then the library will not be updated until the last episode is written. I would like to find a way to get the library updated on each episode one by one, not just hours later..

How should I do this? I could handle season folders differently...

LukePulverenti:
No, let's not start having the library monitor worry about item types. we've gotten this far with what we have now so we don't really need to go overboard. Just keep it simple. If episodes are coming in rapidly they will have to come in continuously every 30 seconds or less in order to cause a delay in getting the new ones scanned in. That's not going to happen very often, and when it does, a delay of the library scan by a couple extra minutes isn't really a problem.

softworkz:
OK, I will do the 'group by folders' approach - which is simple - and see how it goes...

softworkz:
Three new class files may not look like "keep it simple" at first sight, but I think that this is still a good solution because it is more modular and thus easy to understand, control and even modify.

Despite the refactoring, semantics are preserved except:

* Update requests are grouped by parent folders
* Evaluation of 'tempIgnorePaths' is no longer done when queuing paths but instead performed when update paths are being processed. This change avoids the possibility of update paths being lost in certain situations

It's a bit difficult to get the while picture just from viewing the diffs... but please tell me what you think!

softworkz:
* Renamed queue mechanism classes to a more generic wording. In the future, auto-organize folders could be watched in realtime same like library folders. In this case the FileRefreshQueue could be re-used to wait until files are being locked no longer
* Made queue delay configurable
* Added xml comments
* Improved logging

softworkz:
This is just an independent subset of changes included in pull request #1202 anyway. If #1202 would be merged, this PR can be closed.

softworkz:
I removed this change from PR #1202 because it didn't have any dependencies and can be treated separately...

LukePulverenti:
Solved original requirement via 53a4efeff16a0d019ab7bfed63dc9dc5ec4f54bc

PR #1220, opened 2015-10-18
Fixed bug in librarybrowser.js

jmos1277:
Updated getPosterViewItemHtml to fix a bug in which a background image
was not displayed for certain movies in the "Latest Media" section of
the web interface.  More specifically, a movie that was part of a
collection (i.e. has a parent item) AND did not have a Thumb image
would not display any image.  It would attempt to get a Thumb image
from it’s parent (i.e. the collection) but would fail because the
function ApiClient.getThumbImageUrl is seriously broken.

Even if, ApiClient.getThumbImageUrl was NOT broken, it makes more sense
to show a background image for a movie than it does to show the Thumb
of a collection in which it resides.  So the precedence for these two
cases has been swapped.

The ApiClient.getThumbImageUrl function is still broken.  At the time
of this commit is it not clear what situations exercise the
‘options.preferThumb && item.ParentThumbItemId && options.inheritThumb
!== false’ case.  Since that code is broken anyhow, I assume it is not
a very important case.

### DIFF
--- a/MediaBrowser.WebDashboard/dashboard-ui/scripts/librarybrowser.js
+++ b/MediaBrowser.WebDashboard/dashboard-ui/scripts/librarybrowser.js
@@ -1643,14 +1643,6 @@
                     enableImageEnhancers: enableImageEnhancers
                 });
 
-            } else if (options.preferThumb && item.ParentThumbItemId && options.inheritThumb !== false) {
-
-                imgUrl = ApiClient.getThumbImageUrl(item.ParentThumbItemId, {
-                    type: "Thumb",
-                    maxWidth: thumbWidth,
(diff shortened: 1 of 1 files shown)


LukePulverenti:
Thanks. The issue was actually elsewhere.

jmos1277:
Luke, your comment makes it sound like this issue has been fixed.  However, the issue still exists.  I've attached a screenshot of the "Latest Media" section of my Emby Home page to illustrate the behavior.  Both the movies, "Muppets Most Wanted" and "The Pink Panther Strikes Again" have background images.  However, a bug in the librarybrowser.js file prevents those images from being shown.  

screen shot 2016-02-02 at 12 33 44 am

PR #1246, opened 2015-11-01
Fix exception when episode title is null

softworkz:
Sometimes TheTVDb does not have episode. This caused an exception in
EpisodeFileOrganizer


softworkz:
I am wondering if this should be handled at the provider level to return something generic like "Episode 1"...

LukePulverenti:
No, best to let the consumer decide how to handle it.

softworkz:
OK. I didn't do it anyway because the library seems to handle it fine.

PR #1250, opened 2015-11-06
Support for anamorphic transcoding to upscale and maintain available …

paul-777:
Support for anamorphic transcoding to upscale and maintain available detail rather than downscale

Fix issue in library monitor which would fill the log with errors when library folder only had readonly permissions

### DIFF
--- a/MediaBrowser.Api/Playback/BaseStreamingService.cs
+++ b/MediaBrowser.Api/Playback/BaseStreamingService.cs
@@ -1,12 +1,11 @@
-﻿using MediaBrowser.Common.Extensions;
-using MediaBrowser.Common.IO;
+﻿using CommonIO;
+using MediaBrowser.Common.Extensions;
 using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Devices;
 using MediaBrowser.Controller.Dlna;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.MediaEncoding;
-using MediaBrowser.Model.Configuration;
 using MediaBrowser.Model.Dlna;
 using MediaBrowser.Model.Dto;
 using MediaBrowser.Model.Entities;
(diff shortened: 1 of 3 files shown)


LukePulverenti:
The problem with the bitrate upscaling is that it's potentially a paradigm shift for the api. Clients are instructed to supply their max supported values without regard to the content, so they might pass along 100mbps as their max bitrate.

LukePulverenti:
I will make the LibraryMonitor change manually though, thanks. The probing, yes, probably needs to be done but I know it's going to have a temporary paid period of incorrect detection and it's the harsh reactions that really worry me the most.

paul-777:
Hi Luke, 

Upscaling of anamorphic content is the correct thing to do though - down scaling them results in half the pixels and a very poor quality image compared to the intended resolution, and in fact results in 20% less pixels than the anamorphic source which clearly isn't great (based on normal DVD formats). Just because we allow the bitrate to go higher to achieve this doesn't mean it will actually go to the maximum rate given, so even if a client did pass a max of 100Mbps it is very unlikely that any content would upscale to that - you would likely need to find a 70Mbps anamorphic source to cause that to happen. Also anamorphic content is typically restricted to DVDs so even the upscaled version will still have a lower bitrate that an HD stream - which means that if clients are specifying a max bitrate that allows them to work with HD stream then upscaling of anamorphic should be no issue at all.

Actually in my own build I don't limit the output bitrate based on the source bitrate, only by the max specified. (remember this is a max not a target) The reason for this is that when transcoding I am often reading a highly compressed stream and encoding quickly with a lower compression ratio so in order to maintain output quality the bitrate has to rise. With the default restriction in place transcoded bluray rips looked terrible on a big screen. This is was most noticeable when we were triggering level41 -> level40 transcoding of a 1920x1080 stream, the quality drop was huge. The only reason I can see for limiting the bitrate based on the source is if we feel that the max bitrates, resolution etc. requested by clients are not accurate and we need to "help" them with that decision. 

The anamorphic detection now allows for a small amount of distortion without flagging it as anamorphic which should prevent issues with badly encoded streams. If it does get flagged then the distortion would likely be noticeable visually.

LukePulverenti:
I'm going to manually merge the scaling change for when a MaxHeight only is supplied. It looks like we have already made the change for MaxWidth over the course of time while this pull request has been open. The only difference is ours has a *2, then / 2, I believe for rounding.

PR #1268, opened 2015-11-13
Support QSV decoder for LiveTV

babgvant:
ProbeSizeOverride was added because the LiveTV probe doesn't need to be that big to figure out codec.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/MediaInfoRequest.cs
+++ b/MediaBrowser.Controller/MediaEncoding/MediaInfoRequest.cs
@@ -16,6 +16,7 @@ namespace MediaBrowser.Controller.MediaEncoding
         public VideoType VideoType { get; set; }
         public List<string> PlayableStreamFileNames { get; set; }
         public bool ExtractKeyFrameInterval { get; set; }
+        public string ProbeSizeOverride { get; set; }
 
         public MediaInfoRequest()
         {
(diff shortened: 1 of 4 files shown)


LukePulverenti:
Andy I can't merge this in. We already have the code to probe live tv but i had to comment it out right now because it causes multiple streams to be opened and closed with the tuners which can create tuner conflicts. It will have to wait until i have a chance to buffer the stream in Emby server.

babgvant:
Where is the existing probe code?

LukePulverenti:
In BaseTunerHost - AddMediaInfo

babgvant:
The reason why it's causing trouble is probably the same problem fixed by using a small probe size (the override I added).

LukePulverenti:
No I mean the trouble is a different kind of trouble, as in other open streams get terminated, recordings fail, etc. We need to make sure the entire playback session only opens one single stream.

babgvant:
How does probing cause that problem? As long as the original stream isn't terminated HDHR tuner locking shouldn't allow that to happen.

LukePulverenti:
Because ffprobe opens one stream, closes it, then the video player or ffmpeg opens a second one.

babgvant:
That's not how HDHR tuner locking is supposed to work. HTTP streams only hold a lock as long as they are open, when it closes it frees the tuner. The second request would either take the previous (now free) tuner, or another one if it's free. If probing breaks an existing stream, then the tuner isn't locked. Which means the DVR implementation is flawed (cough MythTV), or there's a bug in SD's code.

If the server is overlapping the probe and request, that might cause failures, but it doesn't sound like that's what's happening.

LukePulverenti:
You're right, but their problem becomes our problem so for now I had to disable it.

babgvant:
Have you escalated it to them?

PR #1380, opened 2015-12-29
Various improvements for file and folder browsing

softworkz:
* Fixed broken parent folder API call
* Added icons to distinguish files and folders
* Modified all usages to initialize the dialog to the current value when it is displayed
* Modified some usages to also include hidden folders. Otherwise it wouldn't be possible to browse to a subfolder of the Emby installation (on Windows at least)
* Modified API behaviour to return useful results, even if the selected path is pointing to a file instead of a folder

### DIFF
--- a/MediaBrowser.Api/EnvironmentService.cs
+++ b/MediaBrowser.Api/EnvironmentService.cs
@@ -223,8 +223,17 @@ namespace MediaBrowser.Api
         /// <returns>IEnumerable{FileSystemEntryInfo}.</returns>
         private IEnumerable<FileSystemEntryInfo> GetFileSystemEntries(GetDirectoryContents request)
         {
+            var path = request.Path;
+
+            if (request.IncludeFiles && _fileSystem.FileExists(path))
+            {
+                // In case we are browsing and selecting files, we would want to return the contents 
+                // of the folder containing the current file
+                path = Path.GetDirectoryName(path);
+            }
+
             // using EnumerateFileSystemInfos doesn't handle reparse points (symlinks)
-            var entries = _fileSystem.GetFileSystemEntries(request.Path).Where(i =>
+            var entries = _fileSystem.GetFileSystemEntries(path).Where(i =>
             {
                 if (!request.IncludeHidden && i.Attributes.HasFlag(FileAttributes.Hidden))
                 {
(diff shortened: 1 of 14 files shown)


softworkz:
I added icons to distinguish files and folders. I used 'folder' for folders but 'menu' for files because I couldn't find an appropriate file icon.
We should see if we can find or create a generic file icon for this purpose...

LukePulverenti:
I have fixed the parent path issue at a lower level. what is the effect of returning the parent directory contents when a file is requested? includeHidden I'm just going to default to true so that it doesn't have to be sprinkled everywhere

softworkz:
The "sprinkle" (I'm interpreting this as "changes to many files") would still be required to pre-initialize the dialog with the current setting (12 of the 14 fiile changes are doing just this).
The parent directory contents are required to fill the dialog files list, even if a specific file is selected (and pre-initialized).

Just merge this and you'll see that file dialogs are actually working as one would expect.
(Two scenarious need to be tested: folder selection and file selection. File selection is used for the ssl certificate setting)

LukePulverenti:
I've defaulted it at the server level so we don't even need that param anymore. If you want to do more work then we can leave this open but I won't merge as is because I don't want to have the api return the parent folder file list if you request children of a file. I think what you're trying to do is keep the list visible when the user clicks on a file. That should happen inside the picker. 

Defaulting the paths based on previous value is a great idea - but we have to handle what happens when it doesn't exist, which I think you've handled at the api level, but that should also be in the picker. We should keep the api design independent of the specific use case.
Thanks

softworkz:
I implemented error handling directly in directorybrowser.js, not in apiclient!

As to the other issue. You wrote 
> I don't want to have the api return the parent folder file list if you request children of a file
What do you mean by "children of a file"? A file cannot have children.
The API path is "/Environment/DirectoryContents". It didn't seem unnatural for me if that API method would return all contents of a specified file's directory in case the path parameter points to a file.

That way appeared to be the most simple solution to me. I don't think that a client-only solution can be found in this case, because the client has no means to distinguish between files and folders. Thus, any alternative I can imagine would require the introduction of an additional API method like one of these:

- GetDirectoryContentsOrFileSiblings
(same like my change but as a separate method leaving the original method unchanged)
- IsFile
(to determine if item is file or directory and implement different handling in the client)
- GetFileDirectoryOrDirectoryPath
(returns the containing directory path for files and the input path for folders)

What's your preference?

PR #1381, opened 2015-12-29
Improve dashboard error reporting

softworkz:
Previously, most errors on API calls were just swallowed without further notice.
As a result, operations that had failed to execute, often created the impression of a none-responsive or faulty UI, since there was just no visible reaction.

This commit enables display of error messages for _all_ error situations (except those few 401-errors that were already hardcoded).

We will probably identify some exceptional situations where we don't want to display an error message. 
But I think it is much better to blacklist errors here, since blacklist is small and predictable while a whitelist would be huge and never be complete.

### DIFF
--- a/MediaBrowser.WebDashboard/dashboard-ui/bower_components/emby-apiclient/apiclient.js
+++ b/MediaBrowser.WebDashboard/dashboard-ui/bower_components/emby-apiclient/apiclient.js
@@ -133,11 +133,19 @@
 
         function onFetchFail(url, response) {
 
+            var status = '' + response.status;
+
+            if (response.statusText)
+            {
+                status = status + ': ' + response.statusText;
+            }
+
             Events.trigger(self, 'requestfail', [
             {
                 url: url,
                 status: response.status,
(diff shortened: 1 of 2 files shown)


LukePulverenti:
we used to have this buut we can't do it because there are several places where we don't want the failure to fall into global error handling. ultimately what needs to happen instead is it has to be implemented on each individual screen with customized error messages for each.

softworkz:
OK, but couldn't we implement a generic mechanism for this that can be activated individually?

Maybe this would be better than copying the same code over and over...

Apart from this, I also fixed accessing the response headers.

    response.headers["X-Application-Error-Code"] 

does not work anymore. We need to use

    response.headers.get('X-Application-Error-Code') 

instead

LukePulverenti:
No because sometimes we need to look at the response status code and customize the error message. it's not really possible to have one global error message handler because then you end up showing error messages when it's not needed or appropriate.

softworkz:
Yeah, I understood that.

What I meant is something like a dashboard-function that can be specified as an error handler for individual apiclient calls.
Just to avoid writing "Dashboard.alert()" stuff over and over again...

PR #1382, opened 2015-12-29
ServerConfigurationManager: Check if a specified SSL certificate can be used when configuration is saved

softworkz:
Emby can only use certificates in PFX format which include a private key, and where the private key is not protected with a password.

Checking these conditions early helps to avoid misconfigurations and reduce troubleshooting time...


LukePulverenti:
This is good although I will remove the private key check simply because I don't want to pull cryptography into configuration. thanks.

softworkz:
OK, but then at least the description should state, that it must be a PFX file without password.
We shouldn't let all users go through trial and error and having them finally find this information (PFX must not have password) somewhere in the forums...

PR #1385, opened 2015-12-30
New dashboard function for response error handling

softworkz:
This commit adds a new dashboard function for generic handling of api errors.
A sample implementation is included for dashboardhosting.html


LukePulverenti:
Based on previous discussion, can't do this.

softworkz:
Sorry, this is not what you denied with the previous discussion (= general handling of all exceptions)!

It is what you demanded: Error handling needs to be done individually.
While the dashboard function is generic, it needs to be explicitly specified on each API call.

The other commit is just an example how to use it...

PR #1386, opened 2015-12-31
Fix issue where shows without icons would use the last icon that was displayed

J-Swift:
Because of JS hoisting rules, a stale value was being used for image urls. This remedies that.

Here is an example of the bug in action:

duplicated show icons


LukePulverenti:
Thanks!

PR #1396, opened 2016-01-07
Report Fixes

faush01:
Add Not Favorite to filters
Fix export to include all items


faush01:
I have added IMDB, TMDB and TVDB missing filter options

faush01:
I have used an <option> drop down to show the tri state (All, Yes, No) but really should use a Tri State checkbox. I found the following one:
http://vanderlee.github.io/tristate/
But could not work out how to add it to the project.

PR #1398, opened 2016-01-10
Extended OmdbItemProvider to support series episodes

softworkz:
This new episode provider implementation does not bulk-download or cache
episode data. It is only meant to be a backup source for situations
where media is not recognized by that default provider.


softworkz:
As requested I performed these changes:

* Separated into a separate class because it needs a different ordering from other omdb providers
* Implemented IHasOrder - made sure default order puts it after tvdb
* Updated ServerConfiguration to disable it altogether by default
* String.format calls now use CultureInfo.InvariantCulture

softworkz:
Finally added a migration task.
After adding the new code, the provider became disabled once.
Then I re-enabled the provider and on subsequent start-ups, it didn't get disabled again ==> Just as it is supposed to be...

PR #1407, opened 2016-01-19
Fix lots of it's to its

svenstaro:


LukePulverenti:
Thanks, I will implement manually. We don't want to change all those translation files because the translation service will just change it back. I will make the change in the master version.

PR #1426, opened 2016-01-28
Add support for Trailers at Season level and Episode Level

greenlanternprime:
This will allow the addition of trailer links at the season and episode level, visible in the web interface


greenlanternprime:
This is to allow support for a provider plugin to provide trailers for TV

greenlanternprime:
This is to provide support for future plugin providers to provide trailer urls

PR #1432, opened 2016-02-05
Supersedes #1192: Auto-Organize - Added feature to remember/persist series matching in manual organization dialog

softworkz:
When a filename cannot be auto-matched to an existing series name, the
 organization must be performed manually.
 Unfortunately not just once, but again and again for each episode coming
 in.

This change proposes a simple but solid method to optionally persist the
 matching condition from within the manual organization dialog.
 This approach will make Emby "learn" how to organize files in the future
 without user interaction,


softworkz:
@LukePulverenti recently posted this in the original PR:

> I'm ready to merge this one. Can you:
> •sync to latest dev
> •remove the de language text
> •remove the javascript.json text
> 
> Thanks.

LukePulverenti:
I'm going to merge this but a few more changes are needed:

* The ul data-role=listview should be changed to use the paper list instead of the jquery mobile widget.
* The html page added is pulling in a bunch of jquery mobile dependencies but it doesn't look like they're all being used.
* Lastly, avoid writing any new code that uses jquery, so this 
`$('.extractedName', popup).html(item.ExtractedName);`

should be

`popup.querySelector('.extractedName').innerHTML = item.ExtractedName;`

softworkz:
I did as requested:
- Rebased to latest dev
- Removed the de.json file changes
- There was no modification to the javascript.json file anymore

Some general notes about implementation:
- SmartMatchInfos is a new property of the AutoOrganizeOptions class and not of the TvFileOrganizationOptions class. Even if we don't envision smart match functionality for anything but series episodes at this moment - the time may come (perhaps music organization by artist or something we couldn't think of by now) - or may never come. But the model is ready for this - just in case
- EpisodeFileOrganizer: There are quite a few changes replacing TvFileOrganizationOptions by its parent class AutoOrganizeOptions. This is not only due to the previous topic. It is also a preparation for future Movie-Organization. Common functions from EpisodeFileOrganizer will be refactored to a base class FileOrganizerBase from which EpisodeFileOrganizer and the new MovieFileOrganizer will be derived. This is another second reason to use the more generic AutoOrganizeOptions class instead of TvFileOrganizationOptions

You wrote:
> The ul data-role=listview should be changed to use the paper list instead of the jquery mobile widget

At the time I did that, I tried to find an existing view with similar functionality that I could use as a design prototype. That was "scheduledtask.html". It is still using data-role=listview, today. 
Once you have changed that, I can try to follow...
> The html page added is pulling in a bunch of jquery mobile dependencies but it doesn't look like they're all being used

Are you referring to anything more than "jqmtable,jqmpopup,scripts/autoorganizesmart,scripts/taskbutton,detailtablecss"?
I think "jqmpopup" and "scripts/taskbutton" could probably be removed. Or is it more you are referring to?
> Lastly, avoid writing any new code that uses jquery, so this  $('.extractedName', popup).html(item.ExtractedName); 

I'm rarely writing new code (- in the case of html, css and javascript, at least!). I'm usually copying and adapting existing code to maximize compatibility. If you would name or suggest a prototype implementation, I could adapt the code in this PR to the specified style.

softworkz:
A few more notes:
- the auto-organize task button is not working (i.e. not displaying a progress bar when active). I encountered this problem a while ago and fixed it in my working branch, but I don't remember what it was, right now
- it doesn't work as smooth as it should. Sometimes the target series and target folder is determined correctly, but the file is not moved. When clicking on the dialog button, an "OK/Cancel" dialog is presented. Finishing with OK, actually copies the file
- Even if far from perfect, this is an important and fundamental change, that so so many users have been waiting for a long time!

PR #1474, opened 2016-02-18
Smart Match Activation

softworkz:
A few small changes to make smart match finally work


LukePulverenti:
Merged manually, thanks.

softworkz:
I really wonder why you are doing this...

If the ExtractedSeriesName is null or shorter than 4 characters - why would you want to display the checkbox in such cases? It just doesn't make any sense...

Same issue with the display of the extracted series name. If the user can't see it, he won't be able to troubleshoot Emby against the current smart match configuration.

Why?

PR #1477, opened 2016-02-19
Multiple sources for EmbyTV

jose-pr:
Add channel offset to allow multiple sources with same channel number
but different channels to coexist


LukePulverenti:
Cool, thanks! How do you explain a channel offset?

jose-pr:
It just add an offset to the channel of the source so when emby ask for program data it will  not find it in a incorrect listing source if there is another channel with the same number.

jose-pr:
Found some stuff that needs to be fix.

PR #1482, opened 2016-02-21
Capture lyricist and writer for songs

vicpa:
Thanks Luke!! Here you go


LukePulverenti:
Please make these lowercase:

var Lyricist

vicpa:
Fixed!!

PR #1490, opened 2016-02-23
Autoorganizenewseries

softworkz:
This commit adds some modifications to the itemidentifier component so that it can be used to search for new series in the fileorganizer dialog.

It's not a full-featured implementation for "New Series Creation", though.

This commit just adds a "New Series" button to the fileorganizer dialog, which displays the itemidentifier dialog and adds the result to the series dropdown upon completion.

Backend processing for new series is not yet included here!

### DIFF
--- a/MediaBrowser.WebDashboard/dashboard-ui/components/fileorganizer/fileorganizer.js
+++ b/MediaBrowser.WebDashboard/dashboard-ui/components/fileorganizer/fileorganizer.js
@@ -24,19 +24,19 @@
             context.querySelector('.fldRemember').classList.remove('hide');
         }
 
-        $('.inputFile', context).html(item.OriginalFileName);
+        context.querySelector('.inputFile').value = item.OriginalFileName;
 
-        $('#txtSeason', context).val(item.ExtractedSeasonNumber);
-        $('#txtEpisode', context).val(item.ExtractedEpisodeNumber);
-        $('#txtEndingEpisode', context).val(item.ExtractedEndingEpisodeNumber);
-        $('.extractedName', context).html(item.ExtractedName);
+        context.querySelector('#txtSeason').value = item.ExtractedSeasonNumber;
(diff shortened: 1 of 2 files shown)


LukePulverenti:
it's good but don't write any new code using jquery or jquery deferred. See guest inviter for a replacement example

softworkz:
OK, I'll change that. Should I also change the existing code?

LukePulverenti:
No, not necessary for this pull request. Thanks.

softworkz:
OK, changed new code - does not use jquery selectors anymore and itemidentifier.showFindNew is now using native Promise instead of jquery deferred.
I can also change fileorganizer to move away from jquery later, but I didn't want to overcrowd this PR..

softworkz:
You will need to look at the positioning of the "New Series" button. I used fixed percentage values for element widths, but this may be too small for other languages. I didn't find a way to match the button width to its text content and have the dropdown fill the rest of the total width of the row.

Of course, you will also need to hide the new button afterwards, since the feature is not yet completed..

LukePulverenti:
I merged this in by hand. Great work here, it looks great. The one thing to fix, if you click new series and then cancel by backing out, it resets the dropdown value. What you should do is use the reject handler to signal that the operation was cancelled, and then the auto-organize page can catch the rejection and react accordingly, or probably not react at all.

softworkz:
Actually I did the dropdown reset on purpose to remove any "new Series" entry that was added previously. Yes, it also clears the selection. I though this would be no issue because once you clicked on the "New Series" button, you won't be interested in a previous selection anyway.

Is that OK, or do you still want me to change it?

LukePulverenti:
I would change it because the natural expected behavior when cancelling a dialog is that no changes are made. Thanks.

softworkz:
Updated the branch:
- No changes are made on cancelling
- Replaced jQuery stuff in fileorganizer.js

LukePulverenti:
Perfect, thanks!

PR #1491, opened 2016-02-23
M3u Parser Improvements

jose-pr:
Improvements to the parser and recognizes satip format.


LukePulverenti:
This is getting much more complex now with all of the regular expressions. Can you add some unit tests to the MediaBrowser.Tests project? Just create a few sample files with different variations that way anytime we make a change we can run the unit tests to make sure nothing is affected. thanks.

jose-pr:
I will try to add some test but haven't added any new reg expressions, just look for more fields and check if the title starts with a number and a dot.

LukePulverenti:
Can you explain

"	Make to work with current channel id values"

PR #1501, opened 2016-02-26
AutoOrganize new Series - Backend Implementation

softworkz:
Now it's finally working!!!

### DIFF
--- a/MediaBrowser.Model/FileOrganization/EpisodeFileOrganizationRequest.cs
+++ b/MediaBrowser.Model/FileOrganization/EpisodeFileOrganizationRequest.cs
@@ -1,4 +1,6 @@
-﻿namespace MediaBrowser.Model.FileOrganization
+﻿using System.Collections.Generic;
+
+namespace MediaBrowser.Model.FileOrganization
 {
     public class EpisodeFileOrganizationRequest
     {
@@ -13,5 +15,38 @@
         public int? EndingEpisodeNumber { get; set; }
 
         public bool RememberCorrection { get; set; }
+        public string NewSeriesName { get; set; }
+
+        public string NewSeriesYear { get; set; }
+
+        public string NewSeriesProviderIds { get; set; }
+
+        public string TargetFolder { get; set; }
+
+        public Dictionary<string, string> NewSeriesProviderIdsDictionary
+        {
+            get
+            {
+                var dic = new Dictionary<string, string>();
+
+                if (!string.IsNullOrEmpty(NewSeriesProviderIds))
+                {
+                    var str = NewSeriesProviderIds.Replace("{", "").Replace("}", "").Replace("\"", "");
+
+                    foreach (var item in str.Split(','))
+                    {
+                        var itemArr = item.Split(':');
+                        if (itemArr.Length > 1)
+                        {
+                            var key = itemArr[0].Trim();
+                            var val = itemArr[1].Trim();
+                            dic.Add(key, val);
+                        }
+                    }
+                }
+
+                return dic;
+            }
+        }
     }
 }
--- a/MediaBrowser.Model/FileOrganization/EpisodeFileOrganizationRequest.cs
+++ b/MediaBrowser.Model/FileOrganization/EpisodeFileOrganizationRequest.cs
@@ -21,6 +21,6 @@ namespace MediaBrowser.Model.FileOrganization
 
         public string TargetFolder { get; set; }
 
-        public Dictionary<string, string> NewSeriesProviderIdsDictionary { get; set; }
+        public Dictionary<string, string> NewSeriesProviderIds { get; set; }
     }
 }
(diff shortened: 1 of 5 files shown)


softworkz:
Here are the required changes on the server-side to make the new series feature actually work.
I reduced the changes to a minimum that is required to get it going

There is one thing which is very important to me and that I also included in this change: 
It's **selection of the target folder**. If the user has more than just one folder configured for tv series content, he should be able to choose the root folder where the new series will be created.

So please: Do not drop this out for simplification! It is important for me and will be for other users as well. In the future, we can hide this dropdown by default and only display it when a user has configured multiple folders. (I haven't done that yet, in order to have a consistent testing scenario)

Different issue: I'm not sure about how to compose the folder name for a new series. Currently I'm having that hardcoded as:

seriesFolderName = string.Format("{0} ({1})", seriesFolderName, series.ProductionYear);

For me that's OK but other users may have different opinions..

MediaBrowser.Model/FileOrganization/EpisodeFileOrganizationRequest.cs
```diff
@@ -13,5 +15,38 @@ public class EpisodeFileOrganizationRequest
...
         public bool RememberCorrection { get; set; }
+        public string NewSeriesName { get; set; }
+
+        public string NewSeriesYear { get; set; }
+
+        public string NewSeriesProviderIds { get; set; }
+
+        public string TargetFolder { get; set; }
+
+        public Dictionary<string, string> NewSeriesProviderIdsDictionary
```
LukePulverenti: This is a little bit messy. The model object should be kept clean and have only the strongly typed dictionary. Leave the parsing up at the api level. I don't even think you need that anyway though, just a Dictionary property.

softworkz: I kept the NewSeriesName and NewSeriesYear separate in the API for the following reasons:
- A future implementation (or different client using the API) might want to allow explicit specification of the series folder name. Use cases may be trivial like to have "The Good Wife" instead of "Good Wife" but also language related. A series might have been identified as "Crime Scene Cleaner", but the user wants to use the original name "Der Tatortreiniger"
- IMO an API should provide a reasonable amount of flexibility and not be arbitrarily limited exactly to the use case proposed at the time of creation

If my arguments still didn't convince, I'd like to remark that reducing the signature to nothing but the provider ids would require an additional metadata query at the server, because the current process is like this:
- series = new Series();
- series.Name = request.NewSeriesName;
- series.Path = Path.Combine(request.TargetFolder, seriesFolderName);
- series.ProviderIds = request.NewSeriesProviderIdsDictionary;
- series.RefreshMetadata(refreshOptions, cancellationToken);

So, the name is required before metatada is refreshed.

On the contrary, if we would really pass the provider ids only, there is no 100% guarantee that the query by provider ids will result the exact same item that the user had searched by name in the UI!

softworkz: I guess I misunderstood you. 
You're just referring to having both NewSeriesProviderIds and NewSeriesProviderIdsDictionary, right?

LukePulverenti:
Yes that's all i'm referring to. if you put the provider id's into the post body then they should just deserialize into the strongly typed dictionary automatically.

softworkz:
Apologies, I introduced these duplicate properties for testing and forgot about them since!

Regarding auto-deserialization to a generic Dictionary: I couldn't find an existing case in the code, so I opted for manual deserialization at the API level for now.

MediaBrowser.Server.Implementations/FileOrganization/EpisodeFileOrganizer.cs
```diff
@@ -157,7 +157,43 @@ public Task<FileOrganizationResult> OrganizeEpisodeFile(string path, Cancellatio
         {
             var result = _organizationService.GetResult(request.ResultId);
 
-            var series = (Series)_libraryManager.GetItemById(new Guid(request.SeriesId));
+            Series series = null;
+
+            if (request.NewSeriesProviderIdsDictionary.Count > 0)
```
LukePulverenti: Better but still a bit quirky to have a property suffixed with Dictionary. Can that be removed?

What you have now is going to create complicated post query strings that will be difficult to debug. Make the request object property a Dictionary as well, then switch the client-side post to use a json post body.  See updateServerConfiguration for an example.

softworkz:
I renamed the property NewSeriesProviderIdsDictionary to NewSeriesProviderIds.
I also followed your hint at the updateServerConfiguration method, but there is no dictionary involved in the serialized content (not even to speak of a generic dictionary); furthermore the serialized object is coming from the server, modified at the client and returned back, which is not the case in this modification.
I couldn't find an example in the current API where a Dictionary is involved (neither default nor generic)

LukePulverenti:
Just try it.

softworkz:
Generic or non-generic?

ebr11:
DisplayPrefs have a dictionary in them...

softworkz:
Thanks for the hints!

I really had my doubt's - but as a matter of fact it actually works, so here we go...

LukePulverenti:
Merged - only thing missed was changing the apiclient method to use the post body rather than query string. i've made that change so you'll just need to test to ensure it's still functioning.

softworkz:
OK great, I will check!

softworkz:
I just tested, and it's working fine!

PR #1506, opened 2016-02-29
Collection of fixes for Auto-Organize

softworkz:
This PR contains a collection of fixes that I had left out for simplicity.

It's probably easier to look at the individual commits instead of the summary of changes

I'll provide some more information about the individual commits here:

### DIFF
--- a/MediaBrowser.Server.Implementations/FileOrganization/EpisodeFileOrganizer.cs
+++ b/MediaBrowser.Server.Implementations/FileOrganization/EpisodeFileOrganizer.cs
@@ -621,7 +621,7 @@ namespace MediaBrowser.Server.Implementations.FileOrganization
             {
                 var msg = string.Format("No provider metadata found for {0} season {1} episode {2}", series.Name, seasonNumber, episodeNumber);
                 _logger.Warn(msg);
-                return null;
+                throw new Exception(msg);
             }
 
             var episodeName = episode.Name;
--- a/MediaBrowser.Server.Implementations/FileOrganization/EpisodeFileOrganizer.cs
+++ b/MediaBrowser.Server.Implementations/FileOrganization/EpisodeFileOrganizer.cs
@@ -269,16 +269,15 @@ namespace MediaBrowser.Server.Implementations.FileOrganization
 
             var originalExtractedSeriesString = result.ExtractedName;
 
+            try
+            {
             // Proceed to sort the file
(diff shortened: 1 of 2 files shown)


softworkz:
**Use Exceptions instead of null return values**
Previously some methods were just returning null  or empty values in case of encountered errors; as a consequence, the actual reason for failure was never propagated to the UI which instead just displayed a generic message like "Unable to sort xxx because target path could not be determined."
After this change, the actual reason for failure will be saved to the auto-organize log or displayed in the UI (when completing the organize dialog).

This information is quite important for the user. Examples are "No permission", "Target folder not available", "Disk full", etc..

softworkz:
**Fixed error handling**
- adjusted onApiFailure handlers due to use of fetch API (using e.headers.get instead of e.getResponseHeader)
- if a server-side exception occurs on completion of the correction dialog, the error message is now displayed as a message box in the client (transferred to the client as response header 'X-Application-Error-Code')

softworkz:
**Clean up autoorganizelog.js**
Cosmetic changes only

softworkz:
**Fixed bug in apiclient.js**
With the fetch API response.headers doesnt have an indexer. headers.get must be used

softworkz:
**Corrected quotes in autoorganizelog.js**
More cosmetics...

softworkz:
**EpisodeFileOrganizer.GetNewPath; Throw Exception when metadata cannot be found**
By throwing the Exception the message will added to the StatusMessage or
displayed in the client.

Without this change, the user would see "...because target path could
not be determined", which is not helpful...

softworkz:
**Bugfix for TvFolerOrganizer.cs**
Without this change the logged message is: "Error organizing episode CommonIO.FileSystemMetadata"

softworkz:
Rebased after first commit has been merged into dev

softworkz:
Any objections about these commits? Most of them are highly trivial...

softworkz:
The latest commit fixes a repeated bug in the TvdbSeriesProvider that leads to exceptions like the one in the attached log snippet.

[ProviderBug_Log.txt](https://github.com/MediaBrowser/Emby/files/157697/ProviderBug_Log.txt)

softworkz:
Now I have even removed that one commit that looked complicated.
The remaining changes should be easy to follow now.

softworkz:
Rebased and added new fix

softworkz:
Rebased once again

softworkz:
Rebased and added two additional fixes

softworkz:
Rebased. 4 commits remaining :-)

PR #1507, opened 2016-02-29
Auto-Organize: Async operation and instant feedback UI

softworkz:
**Important: This is a follow-up to PR #1506 and includes that changes**

**The relevant changes for this PR are all included in the last commit _(Auto-Organize: Async operation and instant feedback UI)_**

That commit includes changes to enable and stabilize asyncronous operations in the auto-organize area. Here are the key points:

- The auto-organize correction dialog is now closed (almost) instantly. This means that the user does not have to wait until the file copy/move operation is completed in order to continue (even with local HDs the copy/move process can take several minutes or even much longer with network destination). Thus a user will be able to initiate multiple file sorting operations without having to wait for each previous operation to complete
- This is achieved via a timeout of 2s for the web request at the server side. The purpose of this 2s period is to wait for a possbible exception. If an Exception occurs within the 2s interval, it will be instantly displayed at the client in a message box. Exceptions occuring after the 2s period are quite unlikely, but in that case the log entry will still be displayed in red color and the error message will be displayed after clicking on the item
- This commit also implements locking of files to be organized in order to prevent parallel processing of the same item. In effect, there can be 2 or more manual organization operations active even while the normal auto-organization task is running without causing any problems
- The items that are currently being processed are indicated as such in the log with an orange color and a spinner graphic
- A side effect of this is that other clients showing the auto-organize log at the same time are always up-to-date as well

Even if one would think that the waiting time to organize a single episode should be acceptable (which I think is not..): The situation will get worse when we add movie organization!
Instead of 1 or 2GB for episode we'll have to deal with files up to 10-20GB. Nobody will have enough patience to wait for this....
And: An operation with long completion time will always make users feel that an error has occured.

I think this change is crucial for the future of auto-organize ;-)

### DIFF
--- a/MediaBrowser.Api/Library/FileOrganizationService.cs
+++ b/MediaBrowser.Api/Library/FileOrganizationService.cs
@@ -182,11 +182,9 @@ namespace MediaBrowser.Api.Library
                 TargetFolder = request.TargetFolder
             });
 
-            // For async processing (close dialog early instead of waiting until the file has been copied)
-            //var tasks = new Task[] { task };
-            //Task.WaitAll(tasks, 8000);
-
-            Task.WaitAll(task);
+            // Async processing (close dialog early instead of waiting until the file has been copied)
+            // Wait 2s for exceptions that may occur to have them forwarded to the client for immediate error display
+            task.Wait(2000);
         }
 
         public object Get(GetSmartMatchInfos request)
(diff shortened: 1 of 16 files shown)


softworkz:
Here's a video demonstrating the visual result of this change:

https://www.youtube.com/watch?v=Lq4RGXVh_os

softworkz:
Rebased and consolidated commits.

NEW: Removed jquery from autoorganizelog, autoorganizetv and autoorganizesmart!

softworkz:
Replaced by #2076

PR #1515, opened 2016-03-03
Created new Metadata provider for TheMovieDB: MovieDbEpisodeProvider

softworkz:
Most of the required logic already existed in MovieDbEpisodeImageProvider.
In order to avoid duplicated code, I have moved all common code to a new base class "MovieDbProviderBase" from which MovieDbEpisodeProvider and 
MovieDbEpisodeImageProvider are inheriting now.


softworkz:
I used the unspecific name "MovieDbProviderBase" for the base class, because it could also serve as a base for MovieDbSeasonProvider, MovieDbSeriesProvider and MovieDbSeriesImageProvider in the future. There is still a lot of duplicate code.

softworkz:
I just added the same kind of migration task as I did for the OmdbEpisodeProvider. 

Regarding query performance I would like to note that the new provider is less critical than the OmdbEpisodeProvider because it caches its results to disk (same like TvdbEpisodeProvider).

PR #1540, opened 2016-03-10
Trailers at Episode and Season Level

greenlanternprime:
This will allow the addition of trailer links at the season and episode level, visible in the web interface


greenlanternprime:
This would be useful for using the moviedb api for season promos and epsiode promos

PR #1553, opened 2016-03-17
update lg dlna profile

LukePulverenti:


LukePulverenti:
If you could try this that would be great:

https://github.com/MediaBrowser/Emby/commit/80abd431ec7019f54e63913dc186d63aa245e199

With this change, subtitles are delivered externally to the TV and no longer require burn-in transcoding. I have tested this and verified on my 2015 LG test unit, but I don't know about earlier models. Thanks.

psxlover:
I'm already using "<SubtitleProfile format="srt" method="External" didlMode="" />" in my profile, and it has been working fine.

PR #1576, opened 2016-03-23
TheTVDB Episode Provider: Fixed incorrect fallback to absolute episode numbering

softworkz:


softworkz:
Problem has been described here: http://emby.media/community/index.php?/topic/32725-odd-thetvdb-fetcher-issue/

softworkz:
Cause has been an invalid fallback to absolute episode naming. In the mentioned case:

- Metadata was requested for S19E04
- TheTVDB does not have data for this season
- This made the provider switch to absolute episode numbering and pick data from episode 4

I corrected this behaviour in a way that
- Normal episode numbering is used in when a season number is supplied with the lookup info
- Absolute episode numbering is used when no seaon number is supplied
- Cases are separated. No more fallthrough.

LukePulverenti:
absolute data should be allowed for season 1 though

softworkz:
Why? 
This would only help in situations where the series.zip file would contain "episode-abs-{0}.xml" files only, but no "episode-{0}-{1}.xml" files.
I don't think that such cases exist. TheTvDb seems to always have a "Season 1" at least.

LukePulverenti:
because users might have their files named that way.

softworkz:
Sure, but what I mean is that the "series.zip" from thetvdb will always have both xml files

softworkz:
- A request for S01E01 will be satisfied through "episode-1-1.xml"
- A request for E01 (without season) will be satisfied through "episode-abs-1.xml"

Thus, no need for a fallback in one or the other direction.
For season > 1 fallback to absolute numbering is logically incorrect.

PR #1616, opened 2016-04-02
Bugfix: Timezone-Adjustments for Premiere and Air Dates Lead to Invalid Data

softworkz:
It happens like this:

- My timezone is UTC+2
- Let's look at TvdbEpisodeProvider, Line 410
- DateTime.TryParse creates a date from the string in the xml, assuming
this is a DateTime matching the local computers regional settings
- Thus, the next statement ".ToUniversalTime()" converts the datetime
(assumed as local datetime) to a UTC date
- If the parsed date is "2016-03-30 00:00" the converted date would be
"2016-03-29 22:00"
- This is happening in the provider. So if Auto-Organize or the library
is looking for 29/03 it would find the episode for 30/03 instead

All this is incorrect. PremiereDate or AirDate values should never be adjusted by timezone..

### DIFF
--- a/MediaBrowser.Controller/Providers/BaseItemXmlParser.cs
+++ b/MediaBrowser.Controller/Providers/BaseItemXmlParser.cs
@@ -638,9 +638,9 @@ namespace MediaBrowser.Controller.Providers
                         {
                             DateTime airDate;
 
-                            if (DateTime.TryParseExact(firstAired, "yyyy-MM-dd", CultureInfo.InvariantCulture, DateTimeStyles.AssumeLocal, out airDate) && airDate.Year > 1850)
+                            if (DateTime.TryParseExact(firstAired, "yyyy-MM-dd", CultureInfo.InvariantCulture, DateTimeStyles.None, out airDate) && airDate.Year > 1850)
                             {
-                                item.PremiereDate = airDate.ToUniversalTime();
+                                item.PremiereDate = airDate;
                                 item.ProductionYear = airDate.Year;
                             }
                         }
@@ -657,9 +657,9 @@ namespace MediaBrowser.Controller.Providers
                         {
                             DateTime airDate;
 
-                            if (DateTime.TryParseExact(firstAired, "yyyy-MM-dd", CultureInfo.InvariantCulture, DateTimeStyles.AssumeLocal, out airDate) && airDate.Year > 1850)
+                            if (DateTime.TryParseExact(firstAired, "yyyy-MM-dd", CultureInfo.InvariantCulture, DateTimeStyles.None, out airDate) && airDate.Year > 1850)
                             {
-                                item.EndDate = airDate.ToUniversalTime();
+                                item.EndDate = airDate;
                             }
                         }
 
--- a/MediaBrowser.Providers/Movies/GenericMovieDbInfo.cs
+++ b/MediaBrowser.Providers/Movies/GenericMovieDbInfo.cs
@@ -223,7 +223,7 @@ namespace MediaBrowser.Providers.Movies
                 // These dates are always in this exact format
                 if (DateTime.TryParse(movieData.release_date, _usCulture, DateTimeStyles.None, out r))
                 {
-                    movie.PremiereDate = r.ToUniversalTime();
+                    movie.PremiereDate = r;
                     movie.ProductionYear = movie.PremiereDate.Value.Year;
                 }
             }
--- a/MediaBrowser.Providers/Movies/MovieDbProvider.cs
+++ b/MediaBrowser.Providers/Movies/MovieDbProvider.cs
@@ -94,7 +94,7 @@ namespace MediaBrowser.Providers.Movies
                     // These dates are always in this exact format
                     if (DateTime.TryParse(obj.release_date, _usCulture, DateTimeStyles.None, out r))
                     {
-                        remoteResult.PremiereDate = r.ToUniversalTime();
+                        remoteResult.PremiereDate = r;
                         remoteResult.ProductionYear = remoteResult.PremiereDate.Value.Year;
                     }
                 }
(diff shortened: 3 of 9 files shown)


LukePulverenti:
the ItemsService change should have no effect because apps have been instructed to pass all dates as utc. Can you confirm?

softworkz:
No, this doesn't make any sense for a query either.
When a user enters a date range to limit search results, no adjustment should occur. 
The entered date should be matched/compared to the permiere/air date of the library item as it is officially recorded (and supplied by the providers).

LukePulverenti:
Yea i understand but that's a breaking api change that effects every app, and also, for other types besides episodes, UTC is preferable. In fact, UTC would be preferable for episodes as well if that's what the provider data gave us. Unfortunately what we need is different behavior strictly for episodes and that's just going to create too many trickle down changes of other places that will need adjustment as a result of this change. Alternatively, this can also be fixed in a more surgical manner at the time the by date episodes are imported into the library.

softworkz:
This is not limited to episodes, it is also applicable for movies and other items.

It's just a matter of logical consequence: 
- It is impossible to work with UTC dates. Not for series/episodes, not for movies, because we don't know the time zones for the provided dates
- Thus, we can't convert them to UTC
- The only option is to leave/treat air date and premiere dates as local (local to to the location of premiere)
- Finally, this would lead to the behaviour that the user is expecting anyway and nothing else.
 
Imagine a German user, searching for a certain US item by release date: He reads the release date on a US website and enters this date in the search (he would never convert that date to the European date and time before entering).

**After all, I don't think that this change would break anything, it would just fix all issues.**

Please note that the changes are limited to "PremiereDate" and "AirDate"! For other dates (e.g. like DateAdded) the timezone conversion is absolutely correct!

LukePulverenti:
This is going to require changes everywhere. Apps are instructed that all dates passing in and out of the api are utc. when we save to the database, we convert to utc. This would be perfectly fine if all the provider data was utc, which it actually is for live tv but not all others. Sorry, but right now i would rather just handle the specific issue surgically.

softworkz:
But how?
UTC conversion is not possible.
It affects library and auto-organize (series, seasons, episodes, movies at least).

http://emby.media/community/index.php?/topic/26050-premiered-date-is-showing-up-wrong/page-4?hl=organize
http://emby.media/community/index.php?/topic/33295-auto-organizedaily-shows/

LukePulverenti:
Because even though we are converting to UTC for storage and transfer, we are making sure to convert back when needed for use with the providers, effectively undoing the conversion. It is only the by date episodes that have a breakdown somewhere that needs to be remedied, and I think it is at the point in which they're imported into the library, but I'll know for sure when I actually dig into it.

It is the best of a difficult choice, either a fundamental api change or a quick fix that tackles only the specific issue. The api change would be really difficult due to the generic api that we have. It would force us to tell the app developers, if you're dealing with type X assume UTC, but type Y assume local. that would be really messy. In a perfect world the provider data would be UTC, but it is not so we have to make the best of it.

softworkz:
Why do you think this would require changes everywhere? IMO, the changes are (more or less) complete in this commit.

Please take some time to carefully think through what is happening here right now:

•Local Timezone is UTC+2
•Look at TvdbEpisodeProvider, Line 410
•DateTime.TryParse creates a date from the string in the xml, assuming this is a DateTime matching the local computers regional settings
•Thus, the next statement ".ToUniversalTime()" converts the datetime (assumed as local datetime) from the local datetime to a UTC datetime
•If the parsed date is "2016-03-30 00:00" the converted date would be "2016-03-29 22:00"

This is just plain wrong. It's not one out of two or more feasible alternatives to choose from (and to prefer due to compatiibility). The result is total nonsense and not even close to UTC.

LukePulverenti:
No they are not because you haven't thought about the effect on the other media types where we do have appropriate utc data. In any application with an api it is always preferable to pass all dates across service boundaries as utc, and that is the way we designed the server api. Unfortunately we learned a little bit after the fact about the way the providers handle the dates, but it can still be handled by ensuring that any manipulation on the date objects is undone at the appropriate stages.

Your comment about just plain wrong is not really looking at the whole picture, because there are other areas where that operation is reversed when the original value needs to be manipulated.

softworkz:
OK, I'm getting your point..

How about a new BaseItem property like **PremiereDateIsUtc**? Series, Episodes, Movies would return false, all others true.

Depending on this property, client input could be processed accordingly (e.g. like you said, reverse timezone conversion).

The "plain wrong" statement was targeting the use of **ToUniversalTime ** in the context of the TvdbEpisodeProvider. I'm sorry that I didn't make that clear..

LukePulverenti:
We could have that property on the server if we really wanted, but we need another way of expressing it through the api, either just documentation or something else, just to avoid fattening up the output of every single item. But that really should be treated separately from this, because this is something we can fix on it's own without having to do all of that.

softworkz:
I agree.

Though, if we exclude the changes to **ItemsService.cs**, I think my other changes are still valid, even while we don't yet have some date discrimination indicator, because:

Performing no timezone conversion at all will produce better results than performing that bogus/incorrect conversion like it's done currently.

softworkz:
Today I re-assessed this issue, but I came to the same conclusion as previously:

- Assuming an externally provided DateTime as local and then converting it to UTC is always wrong (unless your local time zone matches the time zone of the external DateTime)
- I already gave an example to illustrate this
- You should also consider this: The premiere or air dates coming from the metadata providers usually don't have a time component anyway! 
- And these properties are displayed without time component in the UI!
- So the only actual effect of calling ".ToUniversalTime()" is that Emby will display incorrect dates (and incorrectly auto-organizes)

I removed the changes to ItemsService and modified BaseItemXmlParser to use DateTimeStyles.None instead of DateTimeStyles.AssumeLocal.
But I think all other changes are mandatory.

PR #1630, opened 2016-04-05
fix transcoding profiles for HDHomeRun Extend (dev)

razzfazz:
Use the correct resolution (960x548) for the internet540 profile.
Actually advertise the internet540 profile.
Remove the obsolete internet720 profile, which is no longer supported on current firmware.


LukePulverenti:
Hi, please either sign the CLA or I can look at implementing similar changes on my own.

razzfazz:
CLA is signed.

PR #1637, opened 2016-04-08
qsv encode live tv

vicpa:
hi luke
been trying to propose this change git is not my friend.

anyway the change to basestreaming turns on the qsv h264 encode for live tv. and if enabled takes out the "-pix_fmt yuv420p " which isn't supported in nv12. Andy/babgvant confirmed both were possible with the current ffmpeg in the forums. Since we have been testing streaming changes thought you might consider this as well

thanks
-vicpa


vicpa:
hi luke,
know you are busy. have you had a chance to look at this. the pull is weird don't know why the commits show. but the changes are correct. 
been running this for a couple on months and works as expected. it is about a 50% reduction in cpu for the transcode. good pr and win for the community. obviously needs wider testing. Maybe you can just make the changes and reject the pull?
then in a couple of months I can bug you about a copy for the audio.

PR #1673, opened 2016-04-21
Dlna: Stop SSDP if disabled, Fixes #1578

### LINKED ISSUE #1578: Unable to disable DLNA
Network is spammed with SSDP discover packets with destination 239.255.255.250:1900, even though all DLNA settings are disabled.

Version: 3.0.5882.0

nfnty:


LukePulverenti:
Thanks. The only problem here is that the automatic port mapper depends on device discovery, so we're going to have to check that setting as well before shutting everything down.

nfnty:
I'll recreate this PR for the dev branch with fixes.

PR #1676, opened 2016-04-23
OmdbProvider: Added local caching

softworkz:
The lack of caching in the OmdbProvider could result in long library
scan time.
This commit adds caching for the OmdbProvider similar to the
MovieDbProvider.
Downloaded metadata is saved locally and only updated if the last
refresh occured 3 or more days before

### DIFF
--- a/MediaBrowser.Providers/Omdb/OmdbItemProvider.cs
+++ b/MediaBrowser.Providers/Omdb/OmdbItemProvider.cs
@@ -1,4 +1,6 @@
-﻿using MediaBrowser.Common.Net;
+﻿using CommonIO;
+using MediaBrowser.Common.Net;
+using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Entities;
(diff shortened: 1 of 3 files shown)


LukePulverenti:
Do any of the omdb image providers depend on the same api urls as the data providers? That's the reason why we have caching for tmdb and tvdb, because we need to use the same data later in the image providers.

softworkz:
No, this is not the case with Omdb. 

But caching will stll speed up subsequent requests for the same item.

softworkz:
You need to consider this:
- A user would typically have a scheduled library scan every 12 or 24 hours or at some point at night
- This is where the provider cache gets filed/update

But additional library scans either triggered by filesystem events or manually by the user will be a lot faster using cached data.

It also seem to me that Omdb is delaying its responses to throttle requests, which makes subsequent querying take even longer.

softworkz:
Just rebased.

Unfortunately, GIT has screwed up the DIFF to OmdbProvider a bit. It's not as complicated as it looks..

PR #1677, opened 2016-04-23
Restrict metadata refresh of missing/virtual episodes

softworkz:
This commit can significantly improe library scan times!
(in my case from 15minutes to 50s)

This improves the following situation:

When a user has a series in his library with lots of episodes (like
thousands), the metadata for ALL series episodes was refreshed during
each library scan. This resulted in very long scan times, even when
using the provider's cached data.

This fix skips metadata refresh for episodes that are NOT present in the
local library and are having an airdate older than 30days (no more
changes likely to occur) during library scans.

But these skipped items would still be refreshed during a full metadata
refresh.

I believe this is an acceptable balance between performance and
accuracy.

### DIFF
--- a/MediaBrowser.Controller/Entities/TV/Series.cs
+++ b/MediaBrowser.Controller/Entities/TV/Series.cs
@@ -255,7 +255,7 @@ namespace MediaBrowser.Controller.Entities.TV
             // Refresh current item
             await RefreshMetadata(refreshOptions, cancellationToken).ConfigureAwait(false);
 
-            // Refresh TV
+            // Refresh seasons
             foreach (var item in seasons)
             {
                 cancellationToken.ThrowIfCancellationRequested();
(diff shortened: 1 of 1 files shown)


LukePulverenti:
I just recently updated omdb and moviedb to skip these. the tvdb image provider already skips these. what further improvements does this bring?

softworkz:
The further improvement is a reduced library scan time.
Sometimes I had about 15min. After the fix I had 50s once. 
Now I tried again and had 1min 25s. Then I reverted the change and tried again: 7min 2s.
This is still 4,5 times faster! (I suppose, it's 15min when tvdb data is old and gets cached again).

The problem here is the TheTvDb provider. Even when using cached data, it is still quite expensive to parse thousands of xml files. If you want to try for yourself, add "The daily show" (3200 episodes) to your library or another series with even more episodes.

PS: I'm running on an 8 core AMD processor, I suppose the effect would be even more drastic on those NAS devices or other low-CPU systems that Emby can run on..

LukePulverenti:
Ok. Only thing is the missing episodes don't currently have a premiere date until they've been refreshed. What you should do is update MissingEpisodeProvider to set that value on new Episodes that it creates.

softworkz:
Another benefit is that the log does no longer grow by 10MB

softworkz:
> Ok. Only thing is the missing episodes don't currently have a premiere date until they've been refreshed. What you should do is update MissingEpisodeProvider to set that value on new Episodes that it creates.

According to my change, an episode that does not have a premiere date gets still refreshed (once) until it has one..

PR #1678, opened 2016-04-23
Use HTTPS for API requests

nfnty:
The following domains are not HTTPS enabled:

www.imdb.com
www.theaudiodb.com
tvlistings.zap2it.com
bot.whatismyipaddress.com

May have missed a few, but at least all Providers are covered.

### DIFF
--- a/MediaBrowser.Providers/BoxSets/MovieDbBoxSetProvider.cs
+++ b/MediaBrowser.Providers/BoxSets/MovieDbBoxSetProvider.cs
@@ -23,7 +23,7 @@ namespace MediaBrowser.Providers.BoxSets
 {
     public class MovieDbBoxSetProvider : IRemoteMetadataProvider<BoxSet, BoxSetInfo>
     {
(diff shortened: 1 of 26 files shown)


nfnty:
`data/library.db` also needs to be updated.

```
UPDATE TypedBaseItems SET data = CAST(REPLACE(CAST(data AS TEXT), "http://www.thetvdb.com", "https://www.thetvdb.com") AS BLOB);
UPDATE TypedBaseItems SET data = CAST(REPLACE(CAST(data AS TEXT), "http://image.tmdb.org", "https://image.tmdb.org") AS BLOB);
UPDATE TypedBaseItems SET data = CAST(REPLACE(CAST(data AS TEXT), "http://assets.fanart.tv", "https://assets.fanart.tv") AS BLOB);
```

LukePulverenti:
Thanks. I think we can just allow users data to get upgraded over time. There isn't a pressing need to do that immediately.

PR #1680, opened 2016-04-23
Dashboard method for response error handling

softworkz:
This commit adds a new dashboard method for generic handling of api errors.
It also changes local implementations to use the dashboard method.

This allows global control of error presentation (e.g. to switch to toast notifications if desired).


softworkz:
Rebased...

PR #1686, opened 2016-04-27
Don't ignore accents in searchs

anderbytes:
**CHANGE 1:**
In 09/Jan/2015 the search code was changed (https://github.com/MediaBrowser/Emby/commit/2e9645c1e06428bf3223b015b0eac90afecef84f) so that accented chars get explicitly "unnacented" using .RemoveDiacritics()  function before submitted to search.

What I propose here is to stop doing this, because foreign users that try to search things that have accents can't do those searches at all:
- Ex: if a movie has "POLÍCIA" in it (police in portuguese), it can't be found writing POLI nor POLÍ in search field.

If there's ANY reason to removediacritics here... please let's discuss this.
If not... please accept this pull request.

ps: I tought about putting some commentaries like " // don't ever removediacritics here, it's bad for foreign users trying to do searches"  near the start of the function. Good or bad idea?

**CHANGE 2:**
I suggested bringing the Trim before the "isnullorwhitespace" check. Because if "spaces" are being typed without real chars, the expection will the thrown in that case.

Thanks

### DIFF
--- a/MediaBrowser.Server.Implementations/Library/SearchEngine.cs
+++ b/MediaBrowser.Server.Implementations/Library/SearchEngine.cs
@@ -86,13 +86,12 @@ namespace MediaBrowser.Server.Implementations.Library
         private Task<IEnumerable<SearchHintInfo>> GetSearchHints(SearchQuery query, User user)
         {
             var searchTerm = query.SearchTerm;
+            searchTerm = searchTerm.Trim();
 
             if (string.IsNullOrWhiteSpace(searchTerm))
             {
                 throw new ArgumentNullException("searchTerm");
             }
-
-            searchTerm = searchTerm.Trim().RemoveDiacritics();
 
             var terms = GetWords(searchTerm);
 
@@ -211,8 +210,6 @@ namespace MediaBrowser.Server.Implementations.Library
             {
                 throw new ArgumentNullException("input");
             }
-
-            input = input.RemoveDiacritics();
 
             if (string.Equals(input, searchInput, StringComparison.OrdinalIgnoreCase))
             {


LukePulverenti:
No, I think it should stay as-is and the database query should be adjusted to return results with and without special chars.

anderbytes:
Luke, the database is not returning items that have accented chars. I can reproduce it easily here... can you?

LukePulverenti:
Understood, and that's what should be changed.

anderbytes:
I don't if that can be done... that would envolve running removediacritics in DB SQL, and data is written in DB with accents.

What do users gain with this removediacritics? English data doesn't have accents... so is that ONLY so people can type wrong "fást" and bring it anyway "fast" ?

anderbytes:
If the SQL query in https://github.com/MediaBrowser/Emby/blob/445eba31a495b96c663245348e000b264105475a/MediaBrowser.Server.Implementations/Persistence/SqliteItemRepository.cs#L1999 could be made in a way that the query also ignores the accents... everything would be OK.

But unfortunately I don't believe that can be made. Making a search with "sqlite diacritics" in google returned nothing useful, I fear.

anderbytes:
I guess I found something .... don't know if that can be implemented in Emby's SQLITE

http://stackoverflow.com/questions/14009261/sqlite-accent-insensitive-search
http://www.sqlite.org/datatype3.html#collation

Resume: using NOCASE collations.

Anyway... the 2 solutions are above:
1. stop removing diacritics in search field
OR
2. modify the query so SQLITE use NOCASE collations

psxlover:
Another option would be to store the title without accents in a field used only for searching.

LukePulverenti:
^ yea that is probably what's going to happen.

anderbytes:
Well... Personally I'm not a big fan of creating redundant fields in DB for things that could be settle with programming but, it's your call.

If you make this transparent to us users, and manage to always keep the accent-less version up-to-date every time the title changes, then OK.

But notice that : will you create redundant fields for anything that is searchable and can have accent? Movies...Episodes...Persons... ?

LukePulverenti:
Resolved for next release.

anderbytes:
Thanks Luke, I'll test again next stable.

PR #1706, opened 2016-05-04
Add additional Game Systems to MediaBrowser.Model.dll

Pengwyns:
Add additional Game Systems to MediaBrowser.Model.dll to support media stubs in GameBrowser for PlayStation 3, PlayStation 4, Wii U, Xbox 360 and Xbox One.


LukePulverenti:
Let's do this a little differently. I'm going to delete this class from the core since it's not needed in the core server. If it's needed by the plugin, just take a copy of it there. Thanks.

PR #1723, opened 2016-05-10
TV Maze Provider for Series and Episodes

softworkz:
This is a new metadata provider implementation for "TV Maze".

It can do Series and Episodes (GetSearchResults as well as GetMetadata) and also implements IItemIdentityProvider for both.

The architecture is very similar to the TvdbSeriesProvider. All series data (including all episodes) is downloaded at once and cached locally.

As a result, this provider should perform very well (no provider queries for single episodes).

This provider delivers cast information at the series level but not at the episode level. Cast information is cached as well (in a separate file).

Image data is not processed at this time (except for the series search results).


softworkz:
Rebased and added IExternalId for TvMaze

greenlanternprime:
This actually ports well, to a provider plug-in... once TVMaze opens up the episode trailers in their API, this could be an awesome resource.

softworkz:
Finally, we have an alternate SeasonProvider!
TvMaze serves season data and the new provider is now able to process it.

I also added a SeasonImageProvider for TV Maze.

greenlanternprime:
After playing with a plug in version of this which I cobbled together, I've noticed it kept getting confused over "the blacklist" other than that it's spot on, would love this in core, on episode details I added the homepage link to the maze episode webpage, after doing that I was able to link the episode trailers to each episode, which is very nice if you have show unaired episodes in your settings, of course this only works in the web client, and only in my custom build, might try a minimised version of my last pull request, and see if that get approved, tv maze is the only source I've found that has loads of trailers at episode level

softworkz:
> it kept getting confused over "the blacklist" other than that it's spot on

No problem on my side. Here's the result from my MetadataViewer plugin:

blacklist_provider_info

softworkz:
> I added the homepage link to the maze episode webpage

I had missed that in the first version, but I added the TV Maze external id classes for Series, Seasons and Episodes, so all links should work

softworkz:
>  after doing that I was able to link the episode trailers to each episode

You can't get trailers through the API at this time. I posted a request on their site to include them:
http://www.tvmaze.com/threads/1446/trailers

greenlanternprime:
I know, I use the episode webpage and the agility pack to parse the page and pull out the YouTube links, I would be better to get it via the api but as a workaround it works really well, but until virtual trailers are enabled for episodes in core it's a moot point.

softworkz:
We can't do it by scraping the html page - especially not for virtual episodes.

Currently, we are loading all episode data for a show with a single api request.
Now, imagine a series with 5000 episodes: We would need to perform 5000 http requests...

greenlanternprime:
I know my scraping test was with a plug-in which only did a very small set, just to see how well it would work, 

The blacklist thing is weird with mine using the identify button I picked blacklist and it resolved back to the blacklist spin off series

softworkz:
I tried again, deleted all provider ids and re-identified "The Blacklist" again, but I couldn't see any problem...

softworkz:
Last rebase of TV Maze stuff

LukePulverenti:
This is on hold. I think TV Maze is best left as a plugin and not a core feature, unless they later add multi-language support.

softworkz:
I think TV Maze is a valuable source with sometimes even better data than omdbapi.
Also, TVMaze provides per-season data, being the only alternative to the MovieDbSeasonProvider!

But no matter if TvMaze is shipped with the server or as a plugin: Emby must allow more detailed control over metadata retrieval. Once this is done, TV Maze should be included anyway IMO.

My suggestion would be: Include TVMaze now in the main server but deactivate it for all new and upgrade installations - except: The TvMazeSeasonProvider should be enabled by default.

But if you want to wrap it into a plugin, go ahead, I'm fine with it. I just don't want to maintain it myself... :-)

softworkz:
Finally created a plugin. The code in the plugin is unchanged, so it could be merged into main anytime,  so I'm closing this.

PR #1754, opened 2016-05-20
Micro-fixes for logging instance creation exceptions

softworkz:


softworkz:
Added two more fixes...

PR #1761, opened 2016-05-21
Proposal for item subtypes

softworkz:
This is just an idea for a subtype implementation


LukePulverenti:
I think it is a good attempt, but in general, the less the system knows about particular types, the more flexible it will be.

For example, adult videos are mostly treated like regular videos, but with some exception. Users might not want them showing up in search, suggestions, etc. instead of making the system aware of adult videos, let's instead allow those exceptions to be configured per library. Everybody wins doing it that way because there are people asking for those things already, and it will be more useful for whatever other sub-types might come along.

softworkz:
OK, if we will have a solution to select metadata providers per library and a setting to prevent content from being mixed with other libraries' content, this would probably even be a better solution. I'm seeing this now!

PR #1762, opened 2016-05-22
OmdbProvider: Drop IHasShortOverview stuff and other improvements

softworkz:
- Changed OmdbProvider to use full plot information
- All the "ShortOverview" stuff is no longer required and has been
removed
- Enhanced OmdbProvider to return actors, director and writer data
- Fixed bug in OmdbEpisodeProvider generating invalid SortName values
due to missing IndexNumber and ParentIndexNumber values

### DIFF
--- a/MediaBrowser.Api/ItemUpdateService.cs
+++ b/MediaBrowser.Api/ItemUpdateService.cs
@@ -292,12 +292,6 @@ namespace MediaBrowser.Api
                 hasTaglines.Taglines = request.Taglines;
             }
 
-            var hasShortOverview = item as IHasShortOverview;
-            if (hasShortOverview != null)
-            {
-                hasShortOverview.ShortOverview = request.ShortOverview;
-            }
-
             var hasKeywords = item as IHasKeywords;
             if (hasKeywords != null)
             {
--- a/MediaBrowser.Api/Reports/Activities/ReportActivitiesBuilder.cs
+++ b/MediaBrowser.Api/Reports/Activities/ReportActivitiesBuilder.cs
@@ -97,7 +97,7 @@ namespace MediaBrowser.Api.Reports
                         HeaderMetadata.Type,
                         HeaderMetadata.Severity,
 						HeaderMetadata.Name,
-                        HeaderMetadata.ShortOverview,
+                        //HeaderMetadata.ShortOverview,
 						HeaderMetadata.Overview,
                         //HeaderMetadata.UserId
                         //HeaderMetadata.Item,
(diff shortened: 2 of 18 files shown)


softworkz:
This change may appear a little extreme at first sight.

Though, all of that "ShortOverview" stuff obviously existed for a single reason only: Overview descriptions from OMDB have been very brief in the past.
But this is no longer true. By querying OMDB with parameter "&plot=full" we are now able to retrieve just normal overview descriptions.
As a consequence, all implementation for "ShortOverview" information is no longer required and could be removed.

LukePulverenti:
From the pure emby standpoint this is possibly good consolidation, but we support nfo generation and kodi supports long and short versions, so the current code is designed to pull in both.

softworkz:
AFAIK there is no place in Emby where the short overview is displayed when the normal overview is available, so why should we keep this info if we aren't using it anywhere?

To me this seems to be a dead end in Emby, once implemented to receive the poor OMDB descriptions.

Regarding kodi: Kodi may have it, but why should we pull that data if aren't using it anyway?

Regarding NFO: In the future, OMDB will always retrieve full descriptions, so why save some "ShortOverview" data to NFO files that usually is empty (unless manually filled)..?

softworkz:
Please think about this just one more time:

- Is there a real-world scenario where a typical Emby user would even notice this change? The ShortOverview property isn't really used anywhere except for writing to NFO or XML
- Are there any Emby users that are using Emby as an application for pulling Kodi data and writing that data to NFO/XML files? Or why should we care about those 5 ;-) users?
- Why write to XML/NFO if the info isn't used anyway after reading back from XML/NFO files?

LukePulverenti:
Well perhaps another question is, what is the pressing reason to change it?

softworkz:
OMDB now has full plot information for all items.
Thus it is no longer required to store this short information to a special "ShortOverview" property (with the purpose to avoid merging the short plot in favor of a longer plot),
Now plot information from all providers can be treated equal!

That's why I thought it is a good time to drop that ShortOverview thing completely as the original reason for it has gone away.

But if you really want to keep it, I will reduce this PR to the actual OMDB changes and will leave the ShortOverview stuff as it is. It's your decision how would you like it?

LukePulverenti:
Besides the overview stuff, has the rest already been merged in via other pull requests?

softworkz:
I just created the new PR #1907 that contains the "other improvements" of this PR only.

softworkz:
But I think, we're still not done with this.

You we're asking, "what's the pressing reason to remove the "ShortOverview" infrastructure?"

I would sum up that reason like this:
- The original cause that lead to the introduction of the "ShortOverview" stuff no longer exists (OMDB now _does_ have overview information of normal length)
- The "ShortOverview" data field is a dead end in Emby because its content isn't used or processed anywhere (Only exception: It's written to xml or nfo files. But when it's read back from such files - it still isn't used, processed or displayed in any way)
- And here's the pressing aspect: With the current implementation we are throwing away ALL overview data that's coming from OMDB for no reason...

softworkz:
Can be closed after reading because content was further split up into #1907 and #1908

PR #1770, opened 2016-05-23
OmdbProvider: Switched to use full plot information and other improvements

softworkz:
- Changed OmdbProvider to use full plot information
- Enhanced OmdbProvider to return actors, director and writer data
- Improved parsing of production year
- Fixed bug in OmdbEpisodeProvider generating invalid SortName values due to missing IndexNumber and ParentIndexNumber values


softworkz:
This is an alternative to PR #1762 without dropping the ShortOverview stuff

softworkz:
Closed because further split up into #1907 and #1908

PR #1783, opened 2016-05-27
Small translation adjustments for pt-BR

anderbytes:
Some smalls fixes.
Most important:
- "Parental" is not a correct grammar in portuguese. "Etária" is the right one.
- Here "TV Network" is called "Emissora", not really called "TV Network"


LukePulverenti:
Thanks, but do you have access to our onesky translation site? you should make the changes there because these edits will just be lost anytime we sync the translations.

LukePulverenti:
If you don't have access, I can give you access. Just PM me the email address that the invitation should be sent to. Thanks.

PR #1784, opened 2016-05-28
Fix display of overview field in metadata editor

softworkz:


softworkz:
Rebased!

softworkz:
All other variants I tried didn't work out, not even the maxlines property (works when changing input in the client but does not work for initial display.
The commited css targets the iron-autogrow-textarea, which supports a max-size according to the documentation while the paper-textarea does not support a max-size (and obviously doesn't even support explizit sizing).
As a result, the input will grow until 90px height and display a scrollbar for larger content.

PR #1785, opened 2016-05-28
Improve result display for itemidentifier

softworkz:
In many cases the titles were displayed truncated (with ellipsis) and it was impossible to decide if an item is the desired one.

This commit changes the card title to wrap over 2 lines.

I also added a tooltip to the search results indicating from which provider a result item came from and which provider ids it has reported


LukePulverenti:
Obsolete due to changes about to go in.

PR #1788, opened 2016-05-29
OmdbProvider: Fix for data fields containing "N/A"

softworkz:
omdbapi.com likes to return "N/A" for fields whithout content.

This can cause problems when metadata results are merged: During merge, the "N/A" value is used, even if other providers (of lower priority) had better data.

The previous implementation had handled this problem for certain fields only.

This commit solves the problem for all fields in a generic way..

### DIFF
--- a/MediaBrowser.Providers/Omdb/OmdbProvider.cs
+++ b/MediaBrowser.Providers/Omdb/OmdbProvider.cs
@@ -4,8 +4,10 @@ using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.Serialization;
 using System;
 using System.Globalization;
+using System.IO;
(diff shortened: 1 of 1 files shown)


LukePulverenti:
Which fields is this currently not working for?

softworkz:
For **Rated** and **Plot** I had actual cases where i got "N/A" values.
The result was that the Overview was "N/A" and the Rating was "N/A", while another provider would have had real information.

softworkz:
For other fields we use .TryParse(), which of course fails for "N/A" entries, but I think a generic replacement of these values is a cleaner solution.

softworkz:
Rebased and reworked without using TypeDescriptor

PR #1789, opened 2016-05-29
Constrain movie suggestiions to current library

softworkz:
Just like "Resume" and "Latest Movies", the suggestions should be
restricted to the current library as well.

(appears like it was intended anyway but probably forgotten?)


LukePulverenti:
Most users don't have massive libraries. At the moment I don't want to do this. Thanks though.

PR #1794, opened 2016-05-31
Remote Trailers support at episode Level

greenlanternprime:
Minimal changes to support linking remote trailers [eg. YouTube] to episodes, to support manual editing or future meta providers [eg. tvmaze one the api includes this info]


greenlanternprime:
Added this as a minimal change, this will provide instant support for trailers on the web interface, other clients would need a bit of changing to support.

LukePulverenti:
Can you submit this on the dev branch? Thanks!

greenlanternprime:
Submitted to the Dev branch...

PR #1795, opened 2016-05-31
TheMovieDB image fetching for now only supports 2-digit lang code (ISO-639-1)

anderbytes:
So far, TheMovieDB only supports 2-digit language code for image fetching, as all images are tagged that way.

So this change allows users to fetch images with better chance of getting his language's, instead of falling back to English every time, trying to fetch the exact country-Region format.

As stated in https://www.themoviedb.org/talk/574dc1c592514112080000a1 , the devs of TheMovieDB believe that eventually they'll have to implement the country-Region options.


anderbytes:
@LukePulverenti , I only wished to do changes to 4 files related to image fetching for MovieDB.

Don't understand why it got more than 30 commits. I Still have to learn more about GitHub.

PR #1796, opened 2016-05-31
Remote Trailers support at episode Level

greenlanternprime:
Minimal changes to support linking remote trailers [eg. YouTube] to episodes, to support manual editing or future meta providers [eg. tvmaze one the api includes this info]


greenlanternprime:
I believe I've submitted this correctly, and closed the incorrect request as i couldn't work out how to ammened the request.

PR #1799, opened 2016-05-31
Adjust preferred language for image searches in TMDB

anderbytes:
TMDB images are unfortunately tagged with 2-digit (ISO-639-1) languages, instead of full Lang-Country format (Ex: pt-BR, en-GB...)

This way... when requests where made with full format, instead of getting correct or language-related results using preferred metadata language, it is falling back to english (en), as TMDB doesn't recognize (until now) Lang-Country full format.

Topic https://www.themoviedb.org/talk/574dc1c592514112080000a1?page=1 in TMDB forum refers to the case.
Devs states that eventually the full format will be supported... we just don't know when. Until then, let's use this method as alternative.

**PS: proposing merge to master because DEV and BETA weren't mergeable.**

### DIFF
--- a/MediaBrowser.Providers/BoxSets/MovieDbBoxSetImageProvider.cs
+++ b/MediaBrowser.Providers/BoxSets/MovieDbBoxSetImageProvider.cs
@@ -53,7 +53,7 @@ namespace MediaBrowser.Providers.BoxSets
 
             if (!string.IsNullOrEmpty(tmdbId))
             {
-                var language = item.GetPreferredMetadataLanguage();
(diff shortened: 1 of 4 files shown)


anderbytes:
Hi @Luke, can you tell me if this pull will happen next master?

As we speaked in a previous topic, I believe now the changes will be made at the correct places: TMDB images provider, not in a global way as you rejected before.

Tnx!

LukePulverenti:
Thanks. Softworkz reviewed and submitted a modified version:
https://github.com/MediaBrowser/Emby/pull/1877

anderbytes:
Cool! Professional coding... 😃

PR #1811, opened 2016-06-04
Add a mini player to the WebDashboard

NickBolles:
I have constantly wished that I could browse my library while still playing the item I am watching. 
This PR should satisfy that wish! 

There could be some issues with screen size and the % based size of the mini player. There may need to be a min-height and min-width on the player, but I am not sure what they should be.

Let me know what you think!


LukePulverenti:
Hi, awesome. We used to have one but eventually abandoned it. Do you have any screenshots you can show? Just fyi, I won't be able to look at merging this in until the video players from the web app and Emby Theater are merged into a shared component. Then at that point I'll be able to look at how a mini player would fit into both apps.

NickBolles:
Here are some screenshots

Normal view, showing the new "picture in picture" icon button
full screen

Mini player in "active" mode (showing controls)
mini player active

Mini player overlapping while browsing content
mini player over content

The mini player also stays when going to the admin dashboard too, which is the best part for me!
mini player over admin dashboard


Another issue I see is that the controls are squished when the mouse is active. But the main use of the mini player will be to not be using the controls...

As you probly see I fixed a few things and added a drop shadow to the mini player to help distinguish it from the content below.

NickBolles:
Oh, there is also an animation when going to and from the mini player that makes it look very slick.

LukePulverenti:
I can't offer an eta, but since you're handy with javascript you could take on projects related to the unification that would help ease my workload and get us there quicker.

NickBolles:
I'd be down! Where do I get started?

LukePulverenti:
This will be painful but you can help get jQuery removed from these two script files:
https://github.com/MediaBrowser/Emby/tree/master/MediaBrowser.WebDashboard/dashboard-ui/components/metadataeditor

Emby Theater does not import jQuery and I'd rather not start doing that. I am working on unifying the listviews and card layouts and once these things are done, playback will be the last major piece requiring unification.

So although these things are not directly related to what you want, they are sitting in line in front of it so any help would go a long way to getting us there faster.

NickBolles:
Awesome! I will work on that this week!

NickBolles:
@LukePulverenti how is this looking. Does it need to be revised because of changes since this? 

I haven't gotten to removing jquery from the referenced file above. Is that still needed?

LukePulverenti:
I've already taken care of the jQuery removal. Merging the playback frameworks between the apps is the last thing remaining.

LukePulverenti:
Unification is done in the dev branch. This feature could be added to the htmlvideoplayer component.

LukePulverenti:
We probably have some things to talk about. For starters, in the Emby web app, the video OSD UI is owned by the app, but in Emby Theater, it's owned by the user selected skin. In other words, the OSD is not attached to the video player, they are completely separate.  (And I see you've noticed that). This approach not only allows for skinning but it also allows for the same OSD to be re-used with alternative video players (windows, samsung, linux, raspberry pi, etc).

So if you pop the video player out and navigate back, you are no longer on the OSD screen anymore. So there are a few possible approaches here:

- Just have the video player own it's own mini OSD for docked mode. This would require some duplication of effort for whatever features end up going into the mini OSD.

- Do a minor rework to the OSD to separate it out from the page and into a component that can be embedded into any element. 

In the latter approach I would imagine something like,

- navigate to videoosd.html, this creates an instance of the OSD and attaches it to the OSD page. When the user navigates out of that page, you could then destroy the OSD instance attached to the page, and the docked container could create a new one. Or you can look for an approach to retain the same OSD instance.

NickBolles:
I follow, but it might take a bit to understand and finish up. I'll check it out a bit today and more next week. Thanks for the tips!

LukePulverenti:
Nick, can you please send me a PM in the emby community, thanks.

NickBolles:
Looks like this is part of Emby Server now! closing. Thanks @LukePulverenti!!

PR #1813, opened 2016-06-05
Use HTTPS for API requests: FanArt: http:// -> https:// (2)

nfnty:


LukePulverenti:
Where is http coming from though? I'd rather keep that knowledge as opposed to just having a catch all.

nfnty:
This is just fixing the ones I missed. I made the same changes in PR #1693 (https://github.com/MediaBrowser/Emby/pull/1693/commits/20c3440a0dd4c99fe676502c3b1219e956f42cec), but only for `FanartMovieImageProvider.cs` The HTTP URL is in the API response which is then transformed into HTTPS.

PR #1829, opened 2016-06-08
Display placeholder background for images while lazy loading (without scripting)

softworkz:
This solution uses img elements with predefined src property for
lazy-loading


LukePulverenti:
This is better but now can you strive to improve it even more and eliminate the need of an extra img + div on every card? Look for a css only solution that doesn't require extra elements. thanks.

softworkz:
**This solution is not using an additional img tag**. It is using an img tag **instead** of the div tag that was previously used. So, there is no extra element being used.

If we don't want to interfere with images that have transparency (eg a transparent border as those created by the cover art plugin), the placeholder background needs to be removed (or switched) just in the moment when the actual image is loaded and about to be displayed.

Since we agreed that we don't want an event based solution, there aren't much options left to perform that kind of image switching.

The key to the solution here is the img tag. The special behaviour of the img tag (and this behaviour is consistent across browsers on all platforms I tested) is this: If the img tag is displaying an image via its src property and the src property is changed from javascript, it keeps displaying the previous image until the new image is loaded and ready for display.

For css-backgrounds this does not work. When the css background is set to an image that is not yet loaded, the background is immediately cleared until the new image is loaded.

I also think that using img tags in an html view to display like 100 images is a more natural way than doing it with 100 div tags using css-baclgrounds. I don't have any evidence for this statement, but I would imagine that at least some browsers may handle 100 img tags with greater efficience than 100 css backgrounds.

LukePulverenti:
We can't do this. While I would like to have a visible placeholder img that has it's src replaced, we need to be able to choose between filling a container with stretching and filling a container without stretching. As of today img does not offer that which is why we use background image css. This can be re-evaluated in the future as newer techniques emerge.

softworkz:
Just a few days ago I wanted to show a few pictures to a friend (android app, mobile network): It was just impossible to do that!
That's caused by two things: 1. missing placeholders; 2. scrolling does not stop at the end of the list but far beyond in the dark instead. I wish I could stand next to you demonstrating that behaviour in real-life...

softworkz:
>  we need to be able to choose between filling a container with stretching and filling a container without stretching.

Photos must never be stretched, so this could at least be applied to photos....

PR #1836, opened 2016-06-11
Channelinterface enhancements

softworkz:
- Added ChannelMediaContentType.Person
- remove override for GetClientTypeName (either globally but at least for MusicAlbum!)
- Fix rendering of MusicAlbums from channels as albums
- css: .itemOverview - preserve line breaks from source string
- If the channel plugin clears all sort fields: Don't fallback to name
  sorting (clearing sort fields means that the plugin doesn't want any
  sorting)
- Added appropriate handlng for photo items
- Added handling for (the new) person items
- Improved handling of MusicAlbum items (setting artists from people
  items)
- Added HomePageUrl field to ChannelItemInfo

### DIFF
--- a/MediaBrowser.Controller/Channels/ChannelItemInfo.cs
+++ b/MediaBrowser.Controller/Channels/ChannelItemInfo.cs
@@ -53,6 +53,8 @@ namespace MediaBrowser.Controller.Channels
 
         public bool IsInfiniteStream { get; set; }
 
+        public string HomePageUrl { get; set; }
+
         public ChannelItemInfo()
         {
             MediaSources = new List<ChannelMediaInfo>();
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -1517,10 +1517,10 @@ namespace MediaBrowser.Controller.Entities
 
         public virtual string GetClientTypeName()
         {
-            if (IsFolder && SourceType == SourceType.Channel && !(this is Channel))
-            {
-                return "ChannelFolderItem";
-            }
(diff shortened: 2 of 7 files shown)


softworkz:
This is a list of proposed enhancements for the handling of channel items.
Please tell me what you think!

MediaBrowser.Controller/Channels/ChannelItemInfo.cs
```diff
@@ -53,6 +53,8 @@ public class ChannelItemInfo : IHasProviderIds
 
         public bool IsInfiniteStream { get; set; }
 
+        public string HomePageUrl { get; set; }
+
         public ChannelItemInfo()
         {
```
LukePulverenti, line 59: This is fine. I can add this independent of this pull request.

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -1517,10 +1517,10 @@ public virtual bool IsFolder
 
         public virtual string GetClientTypeName()
         {
-            if (IsFolder && SourceType == SourceType.Channel && !(this is Channel))
-            {
-                return "ChannelFolderItem";
-            }
+            ////if (IsFolder && SourceType == SourceType.Channel && !(this is Channel))
+            ////{
```
LukePulverenti, line 1521: This is going to negatively impact apps that are looking for ChannelFolderItem in order to apply special considerations. While I would like to remove this, at the moment I don't really want to be dragged into all of the app work that this is going to require.

softworkz, line 1521: OK, but could we make an exception for "MusicAlbum" and "MusicArtist"?

softworkz:
This is obsolete. I'm starting a new PR

PR #1868, opened 2016-06-22
Preserve line breaks in item overview

softworkz:
This would be one way to preserve line breaks in overview descriptions.

Another method would be to insert BR tags during acquisition of the description data...


softworkz:
Most items on SoudCloud have formatted description text:

soundcloud_overview_text

softworkz:
Displaying without line breaks would make these descriptions unreadable...

PR #1870, opened 2016-06-22
Channel item sort must not default to name sorting when all sort fields are cleared

softworkz:
A channel plugin must be able to present items in just the order in which it provides items to the host.

This is the "Default" sort order option. In this case, Emby must not perform sorting by "Name".

Sorting by "Name" is a different option. A user can set this or the channel can set this. But if the "Default" option is selected, items must not be sorted by "SortName"


softworkz:
channelsort

PR #1871, opened 2016-06-22
ChannelManager: Update Name and CommunityRating values when changed

softworkz:


LukePulverenti:
Sorry, this was an accidental merge. We can't do this for now due to channels (trailers) that get internet metadata from the server's metadata functions. I'll have to think about how to allow the channel to specify that.

softworkz:
Is there any BaseItem property that would indicate that it has received data from providers?

softworkz:
- Alternative 1: Only force update of item name when it is a channel folder. (I'm showing child counts in the folder names, these should really reflect "live" data)
- Alternative 2: Allow specfying locked fields. If a field is locked (for provider updates), allow updating through the channel
- Alternative 3: Make the SoundCloud plugin a MetadataProvider to update its own content
- Alternative 4: I could "salt" ChannelItemInfo.Id each time it is queried with random content, to make it appear to be a new item and use the new name. (Probably not a good solution, though)

PR #1878, opened 2016-06-25
DirectoryBrowser: Init dialog with existing value

softworkz:
This change will initialize the directorybrowser dialog with the current value wherever it is being used.

### DIFF
--- a/MediaBrowser.Api/EnvironmentService.cs
+++ b/MediaBrowser.Api/EnvironmentService.cs
@@ -227,8 +227,17 @@ namespace MediaBrowser.Api
         /// <returns>IEnumerable{FileSystemEntryInfo}.</returns>
         private IEnumerable<FileSystemEntryInfo> GetFileSystemEntries(GetDirectoryContents request)
         {
+            var path = request.Path;
+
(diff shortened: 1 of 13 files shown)


LukePulverenti:
If the path has become invalid, then the check on FileExists in GetParentPath could potentially cause a long timeout with a loading spinner, leading to the dreaded reports of "not working". So we need to eliminate that check in order to merge this in. However, that kind of defeats the purpose because then you'd be seeing one thing in the text field, and other in the result list.

softworkz:
I replicated the situation. I could reproduce an empty dialog but no timeouts:

When the path no longer exists, then GetParentPath will fail but DirectoryContents will fail as well.
That leads to an empty dialog (because it hits the error delegate of Promise.All at line 87 in directorybrowser.js).
I couldn't replicate a situation where the spinner remains visible (because on error it is hidden in line 88).

I just commited a simple fix for this situation: In case the init path does not exist, the dialog now just behaves like before: Displaying a list of the root drives.

I tested several scenarios where either the file or directory name or the parent path no longer existed: In all those cases the dialog falls back to the previous behaviour now.

Issue #1146, opened 2015-07-25
Authenticate users using LDAP


mcladyz:
Do you LDAP authentication support planned? Thanks for the great software.

LukePulverenti:
I wouldn't rule it out. A few people have mentioned it, but overall we probably need to see a little more demand for it before investing the time.

Larx:
I'd vote for LDAP, too.

sdfg2:
Me too :-)

Cadair:
+1

cdancette:
+1

abeudin:
+1

Aidenir:
+1

joshuaboniface:
I too am extremely interested in this feature! I see a forum post marking it "DONE" on December 12th; any idea when it will make it to release?

LukePulverenti:
Nothing has been done for this.

jstruebel:
+1 from me on this. Is it possible to do this through a plugin or will it need changes to the core server?

quentinus95:
+1

MarinBernard:
+1

mbantegui:
+1 on this, would love to have this feature.

PoGo606:
+1 Can't wait for it !

spdy:
+1

mescribe0:
+1

williamkray:
ugh, i'm sorry to do this, but +1 as well. github really needs a way to star an issue or something to indicate affected users, etc.

markopolo123:
I'd also love to see this.

Half-Shot:
+1 I wonder what's involved in writing a plugin or something.

untotren:
+1

ben-kuhn:
+1

michaelwfullerton:
Me too. Would be great to see this feature

quentinus95:
Any news on this ? It seems that there is "a little more demand" :smiley:

karbowiak:
+1

jalgroy:
+1

quentinus95:
@LukePulverenti Can you give a feedback about this potential feature ?

LukePulverenti:
[...]

PR #1885, opened 2016-06-29
User Authentication: Enable multiple domain providers for authentication.

jose-pr:
Allow the creation of domain providers plugins which the server can use to get users and groups from and the changes to enable this such as encryption/decryption of passwords, creation of local domain, move the configuration and policy settings from xml to the sql database. My goal after this is the creation of  LDAP authentication and the later on to expand on groups to be able to apply default policy by group to easily manage multiple individuals with similar settings.

Relates to  #1146 
I did the work in a local repo and will be updating this pull request with commits while i clean up my code.


jose-pr:
CRUD operations are working in the users(policy, configuration and server user data) and the local_domain (local directory for user auth) database. Left is how to present the users from different domains in the login screen , which i cant think of how to do at the moment.

jose-pr:
Domain now appears on the Login screen, username doesn't need to be the same as the presented name with the picture.

image

jose-pr:
Little change in the way thing works, before external user where they own account now they are part of a local account that is created the first time a external user logs in.

jose-pr:
Pretty much the base work is done the rest is to create a ldap plugin to test it with ldap and to create an interface in users to be able to set external auth in the server. Perhaps convert convert the connect connection to one provided by the new interface.

LukePulverenti:
Jose - thanks! Just want you to know you're not being left hanging. I will start looking at this soon.

jose-pr:
Will do multiple small pull requests to add features needed and may then reopen this pull based on the other ones.

untotren:
@jose-pr is this still up to date? I want to use this so bad.

PR #1918, opened 2016-07-04
Changed to a valid preset for h264_nvenc

phantom-voltage:
When using h264_nvenc, the preset is set to "high-performance" which isn't a valid preset according to the nvenc encoder. For now it's set to default to get a working version.


phantom-voltage:
For some reason it's not detecting my name when I sign the CLA. Must be something with the hyphen.

PR #1932, opened 2016-07-08
Reduced compiler warnings. No functional changes.

softworkz:


softworkz:
Reworked this by removing questionable changes and changing exception handlers to plan 'catch' statements.

PR #1944, opened 2016-07-16
Fixed typo with the FFmpegSavePathNotFound key.

amnuts:


amnuts:
Sorry; I screwed that up a bit!

PR #1974, opened 2016-07-23
Remove Jquery from metadataeditor.js

NickBolles:
As suggested in #1811

I tested a standard workflow, but I couldn't do too much other testing. Let me know if there is anything else I should change or checkout

### DIFF
--- a/MediaBrowser.WebDashboard/dashboard-ui/components/metadataeditor/metadataeditor.js
+++ b/MediaBrowser.WebDashboard/dashboard-ui/components/metadataeditor/metadataeditor.js
@@ -1,4 +1,4 @@
-﻿define(['itemHelper', 'dialogHelper', 'datetime', 'jQuery', 'emby-checkbox', 'emby-input', 'emby-select', 'listViewStyle', 'emby-textarea', 'emby-button', 'paper-icon-button-light'], function (itemHelper, dialogHelper, datetime, $) {
+﻿define(['itemHelper', 'dialogHelper', 'datetime', 'emby-checkbox', 'emby-input', 'emby-select', 'listViewStyle', 'emby-textarea', 'emby-button', 'paper-icon-button-light'], function (itemHelper, dialogHelper, datetime) {
 
     var currentContext;
     var metadataEditorInfo;
@@ -29,7 +29,7 @@
 
         ApiClient.updateItem(item).then(function () {
 
-            var newContentType = $('#selectContentType', form).val() || '';
+            var newContentType = form.querySelector('#selectContentType').value || '';
 
             if ((metadataEditorInfo.ContentType || '') != newContentType) {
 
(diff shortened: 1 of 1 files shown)


LukePulverenti:
Great, thanks! I will test this later. Just looking at it from a glance, it looks pretty good. The important thing was making sure to handle results of the selector functions, handling the cases of 0, 1, or multiple elements, and it looks like you've done that.

NickBolles:
Yea, that's the worst part about removing jquery! Sounds great!

LukePulverenti:
Just fyi, this has been moved to shared components now, until the two apps merge:
https://github.com/MediaBrowser/emby-webcomponents/tree/master/metadataeditor

If you're free to help continue with this, we also need to find a tree that doesn't depend on jquery (or just roll our own):

https://github.com/MediaBrowser/Emby/blob/master/MediaBrowser.WebDashboard/dashboard-ui/scripts/editorsidebar.js

NickBolles:
What are your thoughts about custom web-components for these types of components?

A long while back when Polymer was in 0.5 dev preview I made a menu that took a javascript object and made a deep menu, not exactly a tree, but close. It would need to be completely re-written for Polymer 1, which I havent looked at in a long time but would be interested in looking at.
http://nickbolles.github.io/nb-menu/

LukePulverenti:
We have been backtracking out of Polymer due to two specific reasons - first, there are too many quirks with their elements when running in browsers that are not based on Chromium. If you look at the web client source history you'll see I was maintaining several of my own fixes within the elements before dropping them.

The second reason is performance. The elements look great but don't perform well in browsers not based on Chromium, and even then, you can feel the overhead of pulling in and processing all of the html imports. It's just a little too heavy for me right now.

I am in favor of pure web components though and you can see some examples here: https://github.com/MediaBrowser/emby-webcomponents/blob/master/emby-input/emby-input.js 

These are much lighter, more minimalistic and more stable. They do require some compromise, for example, you might have to render out

`<label><input /></label>`

instead of just
`<input />`

But to be honest, I am OK with making that kind of trade-off. It is similar to the trade-offs you'd experience if you tried to move from Polymer to material-design-lite.

Additionally keep in mind, Emby Theater is targeting smart tv deployements, and the browsers on these tv's are often very primitive. We've been OK with polyfilling registerElement but back when we were trying to pull in all of Polymer it was just a no-go.

NickBolles:
Awesome! These are almost exactly the reasons I haven't been actively using Polymer. 

I'll research web components some more and take a look at this later this week/early next week. Thanks for letting me help out!

LukePulverenti:
So in terms of where we are with moving away from Polymer, it is completely gone from Theater and the shared components. It is also completely gone from the browsing side of the built-in web app. It is only when you move to the dashboard side that you feel the pain of both Polymer and jQuery getting pulled in. But we have replacements ready for everything that's being used, I've just been doing the actual replacement little by little as it's obviously very tedious.

PR #1982, opened 2016-07-24
Fix: Incorrect use of DateTime.ToUniversalTime() results in incorrect dates returned by metadata providers

softworkz:
This is a revised approach to fix the problems caused by incorrect
timezone adjustments.
- Let's look at MovieDbProvider, Line 97
- DateTime.TryParse creates a date from the string in the xml
- The next statement ".ToUniversalTime()" converts the datetime from
  local a local datetime to a UTC datetime
- Emby-Server1 is in Timezone UTC+3: If the parsed date is "2016-03-30
  01:00" the converted date would be "2016-03-29 22:00"
- Emby-Server2 is in Timezone UTC-9: If the parsed date is "2016-03-30
  01:00" the converted date would be "2016-03-30 10:00"

As a result, Emby-Server1 would store a different item date than
Emby-Server2. This is incorrect because the UTC date (per definition)
must be a constant value that is independent of any time zone.

The previous pull request suggested to remove all UTC conversions but
was turned down because it would break consistency of the API.
This commit follows a different approach: It assumes a constant timezone
(US EST - Eastern Standard Time) for all dates returned from the
affected metadata providers.
The result: Emby will continue to use UTC based dates, but the
conversion is done consistently.

### DIFF
--- a/MediaBrowser.Providers/Manager/ProviderUtils.cs
+++ b/MediaBrowser.Providers/Manager/ProviderUtils.cs
@@ -236,6 +236,39 @@ namespace MediaBrowser.Providers.Manager
             }
         }
 
+        public static TimeZoneInfo GetUsEasternTimeZoneInfo()
+        {
+            TimeZoneInfo easternZone = null;
+
+            try
+            {
+                // Windows
+                easternZone = TimeZoneInfo.FindSystemTimeZoneById("Eastern Standard Time");
+            }
+            catch (TimeZoneNotFoundException)
+            {
+            }
+
+            if (easternZone == null)
+            {
+                try
+                {
+                    // Mono/Linux
+                    easternZone = TimeZoneInfo.FindSystemTimeZoneById("America/New_York");
+                }
+                catch (TimeZoneNotFoundException)
+                {
+                }
+            }
+
+            if (easternZone == null)
+            {
+                easternZone = TimeZoneInfo.CreateCustomTimeZone("Custom_EST", TimeSpan.FromHours(-5), "Custom_EST", "Custom_EST");
+            }
+
+            return easternZone;
+        }
+
         private static void MergeShortOverview(BaseItem source, BaseItem target, List<MetadataFields> lockedFields, bool replaceData)
         {
             var sourceHasShortOverview = source as IHasShortOverview;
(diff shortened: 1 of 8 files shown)


drewfreyling:
What are the formats we get back from the metadata providers? Couldn't we just datetimeoffset.parse them (and then UTC converting them) instead of assuming US EST?

LukePulverenti:
Back when we first started we decided all dates would be stored as UTC as an idealistic approach, however, for some metadata fields or providers this has turned out to now always be practical due to the inability to predict the format of the incoming data. We currently do some gymnastics to get around it without too many issues, but at some point an overhaul probably should happen so that whenever it's appropriate we can just keep the source data untouched and not try to do anything with it. However until that time I think it's best to not merge anything like this.

softworkz:
> What are the formats we get back from the metadata providers? Couldn't we just datetimeoffset.parse them (and then UTC converting them) instead of assuming US EST?

Most of the providers don't specify a time base.

But DateTime.Parse assumes a datetime to be local. Converting that to UTC will always result in a value that is totally wrong.

softworkz:
> However until that time I think it's best to not merge anything like this.

I don't understand this. What we are currently doing is complete nonsense.

As mentioned earlier another alternative (which is very easy to implement) would be to eliminate all those UTC conversions for those metadata fields that are just meant to specify a date without time component. I mean, there's absolutely no point in converting a date (without time) to another timezone.

For instance, if my birthday was on 5.6.2016, it's not "5.6.2016 20:00h" in Middle East Asia or "4.6.2016 00:40h" in the US. It's always just 5.6.2016, no matter what time zone you're in!

SludgeVohaul:
Same for photo metadata in the DB (TypedBaseItems.data DateModified) - the imported photo's EXIF DateTime tag is converted to UTC based on the server's location.
As the EXIF specification does not define any timezone information, I think the timestamp should not be modified, or the UTC should be based on the timezone computed from the eventually present Lat/Long data (and  don't forget the eventual daylight saving time). 
After all this has been done, one finally could admire the beautiful sunrise picture taken at 23:00:00 UTC...

softworkz:
> After all this has been done, one finally could admire the beautiful sunrise picture taken at 23:00:00 UTC...

I thought I had already fired all related arguments regarding this issue, but this is something I hadn't thought about. Thanks for pointing this out.
I still got some hope that this will be finally addressed...

schumi2004:
Is this issue resolved in release 3.1.0?
I tried to apply this patch but it gave me a compile error related to ConvertTimeToUtc

PR #1984, opened 2016-07-26
Add StringUsageReporter

softworkz:
StringUsageReporter reports usages of localization strings in the web
application and can also list all unused strings


softworkz:
**StringUsageReporter Class**

This class contains tests for reporting the usage of localization string tokens in the dashboard-ui or similar.

**Remarks**

Run one of the two tests using Visual Studio's "Test Explorer":

-   ReportStringUsage()

-   ReportUnusedStrings()

On successful run, the bottom section of the test explorer will contain a link "Output". This link will open the test results, displaying the trace and two attachment links. One link will open the output folder, the other link will open the output xml file.

The output xml file contains a stylesheet link to render the results as html. How that works depends on the default application configured for XML files:

-   Visual Studio
Will open in XML source view. To view the html result, click menu 'XML' =&gt; 'Start XSLT without debugging'

-   Internet Explorer
XSL transform will be applied automatically.

-   Firefox
XSL transform will be applied automatically.

-   Chrome
Does not work. Chrome is unable/unwilling to apply xslt transforms from local files.

LukePulverenti:
Fantastic, thanks!

PR #2005, opened 2016-07-31
Health reporting for Emby

softworkz:


MediaBrowser.Controller/Health/HealthMessage.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        /// <summary>
+        /// Gets the type of health message.
+        /// </summary>
+        public string MessageType
+        {
+            get
+            {
+                return _messageType;
```
drewfreyling, line 72: Why don't you use auto properties here?

softworkz, line 72: Because those values are immutable and auto-properties are unable to express this.

drewfreyling, line 72: You can get the same value out of `public string MessageType {get; private set;}`

softworkz, line 72: No. This restricts access to the setter only, but doesn't make the values immutable.

drewfreyling, line 72: Ah yes sorry you are right. I guess we could use expression bodies here if the CI supports C# 6 - `public string MessageType  => _messageType;`

softworkz, line 72: Immutability is not a functional requirement. Declaring those fields as readonly may affect the compilation but probably with little to none benefit regarding performance.
The primary reason for declaring those fields as readonly is to improve clarity of code and make it easy to understand.  We could find a number of fancy alternatives, but none of them would meet this requirement any better...

softworkz:
Adjusted to removal of paper-fab

drewfreyling:
What is happening with this PR? @softworkz it has merge conflicts.

PR #2025, opened 2016-08-05
Async stream handling: Use interface instead of Func<Stream,Task>

softworkz:
No functional changes!!


softworkz:
Fixed sync conflicts.

softworkz:
Actuallly this is nothing more than a slight coding style improvement...
(but will make streaming implementations much more transparent)

PR #2029, opened 2016-08-06
Fix for bad DirectoryService.GetFile performance

softworkz:


softworkz:
Background:
- Enumerating files in a folder is a very expensive operation
- Opposed to this, accessing a specific file (or its properties) is an extremely cheap operation
- Purpose of the DirectoryService was obviously to avoid repeated file enumerations in a specific folder by caching the results

The implementation of the DirectoryService got one step too far, though:

To get information about a single file, it enumerated all content of the file's parent folder (which is expensive) instead of just returning the properties of that single file (which is cheap).

This fix changes the behaviour and will always directly determine the file information from the file system.

LukePulverenti:
Thanks. Already made the same change.

softworkz:
Hehe, OK!

________________________________________

PR #2033, opened 2016-08-07
Add support for VAAPI hardware acceleration

marmay:
Since ffmpeg 3.1 VAAPI hardware acceleration is supported. This commit adds
support to make use of VAAPI for transcoding. This includes hardware decoding,
scaling and encoding.

Because VAAPI hardware scaling is a bit limited, some parts of scaling were
redone in emby, even though they are supported by the software scaler of
ffmpeg.

As far as I know, VAAPI is a Linux thing. Adavantages over QSV are that it works out of the box if ffmpeg is recent enough (>= 3.1). My experience with this patch is great so far. The part that probably requires discussion is the rewrite of the scaling logic.

I am on holiday for a week now. I wanted to share this though, in order to avoid that someone else spends their valuable time on this. I'll bring this into proper shape for inclusion after my holiday, so sorry, if responding to your criticism takes some time.

### DIFF
--- a/MediaBrowser.Api/Playback/BaseStreamingService.cs
+++ b/MediaBrowser.Api/Playback/BaseStreamingService.cs
@@ -309,6 +309,10 @@ namespace MediaBrowser.Api.Playback
                 {
                     return GetAvailableEncoder("h264_omx", defaultEncoder);
                 }
+                if (string.Equals(hwType, "vaapi", StringComparison.OrdinalIgnoreCase))
+                {
+                    return GetAvailableEncoder("h264_vaapi", defaultEncoder);
+                }
             }
 
             return defaultEncoder;
@@ -422,7 +426,8 @@ namespace MediaBrowser.Api.Playback
 
             if (!string.IsNullOrEmpty(state.VideoRequest.Profile))
             {
-                if (!string.Equals(videoCodec, "h264_omx", StringComparison.OrdinalIgnoreCase))
+                if (!string.Equals(videoCodec, "h264_omx", StringComparison.OrdinalIgnoreCase) &&
+                    !string.Equals(videoCodec, "h264_vaapi", StringComparison.OrdinalIgnoreCase))
                 {
                     // not supported by h264_omx
                     param += " -profile:v " + state.VideoRequest.Profile;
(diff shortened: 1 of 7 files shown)


LukePulverenti:
Thanks! My only questions/concerns are the changes to resolution scaling. For now, I've taken I've merged in the quote fixes by hand, relating to forced key frames, because I vaguely recall an issue around that. Thanks.

marmay:
Regarding the scaling, I made some fundamental assumptions in the code:

* `state.VideoStream.Width` and `state.VideoStream.Height` contain the actual resolution of the video stream, so the values are not rounded to some resolutions that makes sense (e.g. 1920x1080) and do not take the aspect ratio into account.
* The display aspect ratio is passed through into the output container.

I do not know, whether those assumptions are correct. If they are not, then my code most likely breaks stuff.

If those assumptions are correct, however, I think it should do the right thing, which in my opinion is: If the requested output size or a maximum size are set, then set the output size to `min( requested output size, maximum size )`. If those values are different from the input size, add a scaling filter. One could argue, whether it would be better to set the output size to `min( input size, requested output size, maximum size )`, i.e. server does downsampling, but clients do the upsampling.

LukePulverenti:
I think the problem with the scaling here is going to be anamorphic pixels, therefore I'm going to isolate your changes to vaapi. I think users are going to report incorrect scaling but I guess we'll see what happens.

marmay:
Sounds reasonable. I can isolate the changes and rebase the commit, if you want to?

LukePulverenti:
i already did. what is the vaapi device, and is that field mandatory?

marmay:
The field is mandatory, but I think that :0 is a sane default value, but I did most testing with /dev/dri/renderD128.

According to ffmpeg documentation, this is the render node that is used for hardware acceleration. For dual video card systems, it may make sense to use :1 or something else instead.

I think /dev/dri/renderD128 depends on the precise hardware that is used, i.e. I found /dev/dri/renderD64 as well, but :0 should be fairly portable.

LukePulverenti:
Ok, thanks! I will make that the default and add this info as help text on the setup screen.

marmay:
Ah, sorry, I misread the documentation.

:0 refers to the display, i.e. if the server runs without an X server, then this does not work.

/dev/dri/card0 seems to be the best default value.

Btw, thanks for making this such a nice experience. I hope I'll find some time for more pull requests that are not as much work for you as this one. :-)

PR #2040, opened 2016-08-08
FileRefresher.IsFileLocked: Don't log Exceptions in case of no write permission

softworkz:


LukePulverenti:
This is the same as the catch (IOException below it). Whether the file is locked or whether there is no write access, they will both produce UnauthorizedAccessException, so that means this change will nullify all of the checks of whether a file is locked or not.

softworkz:
I doubt that. IOException and UnauthorizedAccessException don't inherit from each other.
Perhaps you mean that UnauthorizedAccessException would be caught by the generic catch anyway.
But the only goal was to avoid writing the log full of detailed Exceptions (Logger.ErrorException()) in case Emby does not have write permissions on a library...

LukePulverenti:
Ok, my memory is fuzzy so i need to wait and see examples of exceptions from files being locked before merging this.

softworkz:
Recently a user posted such a file, I'll try to find it..

softworkz:
Found it: http://emby.media/community/index.php?/topic/37453-high-cpu-and-ram-usage/

It's the link in the first post

LukePulverenti:
Sorry I meant a log from a locked file situation

softworkz:
I just tried this:

I added a file to the library and locked it with: http://www.jensscheffler.de/filelocker

I debugged this and got an IOException. The following line was logged:
2016-08-13 08:32:27.2452 Debug LibraryMonitor: E:\Movies\TestMovie\TestMovie1.avi is locked.

So, a locked file does not cause an UnauthorizedAccessException - qed.

But anyway: This commit is not about changing behaviour, it's just about reducing log lines.

PR #2051, opened 2016-08-13
Fix server to enable client caching of static media files served as partial content

softworkz:


softworkz:
This is the result of intense investigations about Chrome's behaviour regarding buffering of media content. Whatever scenario I tried, with different content files, different player frontends, different server behaviour (supporting range requests or not), or various HTTP response header combinations:

On any seek operation (no matter if large or minimal), Chrome always initiated a new http request and never used any local cached data.

Long story short: As opposed to most other browsers, chrome does not implement specific buffering for media content. Instead it just employs the default http caching mechanism. Thus, the solution was to add appropriate caching headers to the range responses.

Note: This commit only addresses serving of static files to the clients: Either files that match the client's device profile or pre-transcoded files (e.g. created from the FolderSync plugin).

softworkz:
Finally, Chrome is using cached data on seeking when available:

chrome_media_caching

LukePulverenti:
This is currently working as expected.  The caching headers being applied are conditional (weak), due to the fact that the url has nothing but library Id's. Therefore if we switch to strong caching, there is currently nothing in place to bust the cache if the source file changes. If you look at the images api, we put a caching value in the url, basically the equivalent of an Etag. This allows clients to cache forever based on that url because if the image changes, so too will the url. In order to switch to strong caching, we would need some kind of an api change such as that.

LukePulverenti:
https://github.com/MediaBrowser/Emby/commit/dcf3509ea28e1bb0cdc3ada16432cfb50b08aac4

softworkz:
This solution used the cache key as ETag and the cache key is built from file path + last modified date, so it is a strong key!

I will test your other commit, but my commit is more universal as it doesn't need any client scripting.

I have investigated quite a few other servers streaming html video and they were all sending the http response headers in the way I did in this commit!

LukePulverenti:
See: "there is currently nothing in place to bust the cache if the source file changes". That's what I added in my commit which then allows the normal caching mechanisms to be used. What you did in range request writer was just repeating the same response headers that HttpResultFactory already writes.

softworkz:
No, these headers (Cache-Control, Last-Modified, Expires and ETag) were not written before I added them in RangeRequest writer.

LukePulverenti:
Yes, some of them were, with conditional (weak) caching. If you look at the control flow of the caching headers you see the difference between conditional and unconditional is the presence of a cache duration. The consumer up in the api layer was not supplying a cache duration (and rightfully so), therefore it was only getting conditional caching based on last modified and etag. I updated the api layer to pass in a long cache duration which now steps it up to unconditional and allows the client to cache forever. But before this could be done, we first had to devise a way to change the url in the file changes because that would be the only way to force the client to download it again.

softworkz:
Correction: They were ALL missing when I was working on streaming live transcoded files.

What was missing for static files served via RangeRequestWriter was the "ETag" and Chrome only started caching after supplying the ETag.

softworkz:
> But before this could be done, we first had to devise a way to change the url in the file changes because that would be the only way to force the client to download it again.

I noticed that chrome always performs a remote request when playing video, even if it has cached it, so I thought this would be enough to detect a changed ETag in the response..

LukePulverenti:
Ok, then yea for transcoded output that would be expected because we don't know what the final content length is going to be. The range request writer doesn't even get used for that. That's the progressive stream writer up in the api layer.

softworkz:
Actually it's my new "ProgressiveRangeWriter" ;-)

LukePulverenti:
Yes that is conditional caching which was already there. That requires the client to come back and check if it has changed via conditional requests. Unconditional says you can keep your copy until X expiration date without ever having to come back to see if it has changed. The only way to bust that is with a new url altogether. Granted with a large video file this is unlikely to ever be a problem, but it is possible for music.

softworkz:
Yeah, I understand!

softworkz:
I tried with your changes only and it does not work. 
Chrome is always requesting from the server:

chrome_media_caching2

LukePulverenti:
ok, what is the exact difference?

softworkz:
Request URL:http://home1:8096/emby/Videos/0871f994fbc9f219512ebecc2715786d/stream.mp4?Static=true&mediaSourceId=7DE6E4164BAC7C92B532D04E9B2C3D07&deviceId=b3ba207e352be3d830accaf8eb77c94f&api_key=9ac8250600e4431f8b2ef4e69f2c1927
Request Method:GET
Status Code:206 PartialContent
Remote Address:192.168.20.15:8096

Response Headers

Accept-Ranges:bytes
Access-Control-Allow-Headers:Content-Type, Authorization, Range, X-MediaBrowser-Token, X-Emby-Authorization
Access-Control-Allow-Methods:GET, POST, PUT, DELETE, PATCH, OPTIONS
Access-Control-Allow-Origin:*
Cache-Control:public
Content-Length:3145063846
Content-Range:bytes 29720576-3174784421/3174784422
Content-Type:video/mp4
contentFeatures.dlna.org:DLNA.ORG_PN=AVC_MP4_MP_HD_1080i_AAC;DLNA.ORG_OP=01;DLNA.ORG_CI=0;DLNA.ORG_FLAGS=01D00000000000000000000000000000
Date:Sat, 13 Aug 2016 23:08:31 GMT
ETag:"15f846a09b672c71dfaae85269a6aed2"
realTimeInfo.dlna.org:DLNA.ORG_TLAG=*
Server:Mono-HTTPAPI/1.1, UPnP/1.0 DLNADOC/1.50
transferMode.dlna.org:Streaming
Vary:Accept-Encoding
X-UA-Compatible:IE=Edge

Request Headers

Accept:*/*
Accept-Encoding:identity;q=1, *;q=0
Accept-Language:en-US,en;q=0.8
Connection:keep-alive
Host:home1:8096
Range:bytes=29720576-
Referer:http://home1:8096/web/itemdetails.html?id=0871f994fbc9f219512ebecc2715786d
User-Agent:Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36
Query String Parameters
view source
view URL encoded
Static:true
mediaSourceId:7DE6E4164BAC7C92B532D04E9B2C3D07
deviceId:b3ba207e352be3d830accaf8eb77c94f
api_key:9ac8250600e4431f8b2ef4e69f2c1927

softworkz:
I had Last-Modified and Expires header
and I had max-age in Cache-Control header

LukePulverenti:
grab the latest code again. that looks like it's missing something.

softworkz:
Yeah, the Tag is missing in the request. But the file is coming from the FolderSync (Dynamic media source), maybe that's the reason..

softworkz:
The media URL is just:

http://home1:8096/emby/Videos/0871f994fbc9f219512ebecc2715786d/stream.mp4?Static=true&mediaSourceId=7DE6E4164BAC7C92B532D04E9B2C3D07&deviceId=b3ba207e352be3d830accaf8eb77c94f&api_key=9ac8250600e4431f8b2ef4e69f2c1927

LukePulverenti:
Ok I would just try simpler tests for now with regular library media. MediaSourceInfo has a new ETag property and we're going to have to make sure that is filled out wherever appropriate. For now I have not done that, I've only just made sure it's there for normal library media. And there will also be cases of remote media where it should just be left null.

softworkz:
It's hard to find a regular library video that is served statically.
Even if I disallow transcoding, they are all running through ffmpeg (either ffmpeg-directstream or ffmpeg-remux). How can I find a file that gets streamed statically or configure Emby to do so?

softworkz:
OK, after adding the ETag to the synced MediaSourceInfo I can confirm that caching works now!

In SyncedMediaSourceProvider I changed the following method:

```
        private void SetStaticMediaSourceInfo(LocalItem item, MediaSourceInfo mediaSource)
        {
            mediaSource.Id = item.Id;
            mediaSource.SupportsTranscoding = false;
            if (mediaSource.Protocol == Model.MediaInfo.MediaProtocol.File)
            {
                mediaSource.ETag = item.Id;
            }
        }
```

softworkz:
Another correction: 

HttpResultFactory.AddCachingHeaders

The header must be named "Last-Modified" instead of "LastModified"

PR #2058, opened 2016-08-14
EpisodeFileOrganizer: Improve error handling

softworkz:
Previously some methods were just returning null or empty values in case of encountered errors; as a consequence, the actual reason for failure was never written to the auto-organize log.
Instead, only a generic message like "Unable to sort xxx because target path could not be determined." was displayed.

After this change, the actual reason for failure will be saved to the auto-organize log or displayed in the UI (when completing the organize
dialog).

This information is very important for the user. Examples are "No permission", "Target folder not available", "Disk full", etc..


softworkz:
I'm having these changes in my private branch since a long time and users will benefit from having precise information about the actual reason for failure.
This is also mandatory when we want to propagate auto-organize errors via the notification system (a planned feature if I remember correctly)

PR #2073, opened 2016-08-18
Hide unused DLNA option

softworkz:
This option does not seem to be used anywhere


LukePulverenti:
Technically it is actually used everywhere since byte range requests for transcoding are always ignored. It's just the opposite scenario that is not supported. Since the toggle doesn't work we can hide it.

softworkz:
Sure, but when we start experimenting with range requests, this could raise wrong expectations, so I thought: better not show that.  (but we don't delete it - maybe some day things change...)

LukePulverenti:
It has to remain. There is one device that needs it, and others that won't be able to function with it. So it has to be controllable either way.

softworkz:
>  (but we don't delete it - maybe some day things change...)

Sorry, I meant to say the same: we should _not_ delete it!

PR #2075, opened 2016-08-19
Auto-Organize: Async operation and instant feedback UI (reworked)

softworkz:
This commit includes changes to enable and stabilize asyncronous
operation in the auto-organize area. Here are the key points:

- The auto-organize correction dialog is now closed (almost) instantly.
This means that the user does not have to wait until the file copy/move
operation is completed in order to continue. (even with local HDs the
copy/move process can take several minutes or even much longer with
network destination).
- This commit also implements locking of files to be organized in order
to prevent parallel processing of the same item. In effect, there can be
2 or more manual organization operations active even while the normal
auto-organization task is running without causing any problems
- The items that are currently being processed are indicated as such in
the log with an orange color and a spinner graphic
- A side effect of this is that other clients showing the auto-organize
log at the same time are always up-to-date as well


softworkz:
Something is wrong here...

PR #2076, opened 2016-08-19
Auto-Organize: Async operation and instant feedback UI (reworked)

softworkz:
This commit includes changes to enable and stabilize asyncronous
operation in the auto-organize area. Here are the key points:

- The auto-organize correction dialog is now closed (almost) instantly.
This means that the user does not have to wait until the file copy/move
operation is completed in order to continue. (even with local HDs the
copy/move process can take several minutes or even much longer with
network destination).
- This commit also implements locking of files to be organized in order
to prevent parallel processing of the same item. In effect, there can be
2 or more manual organization operations active even while the normal
auto-organization task is running without causing any problems
- The items that are currently being processed are indicated as such in
the log with an orange color and a spinner graphic
- A side effect of this is that other clients showing the auto-organize
log at the same time are always up-to-date as well


softworkz:
Replaces #1507

softworkz:
## Testing the new feature
Here are some procedures for testing the new capabilities:

Hint: For more fun, open two or more browsers displaying the auto-organize log simultaneously

### Test Realtime progress display
- Put one or more files into the auto-organize folder (from which you know that they are identified correctly)
- Start the Organize-Task
- Watch how all new files are added to the top of the list in real-time, displayed as "In-Progress" while they are copied

### Test instant dialog completion
- Put one or more files into the auto-organize folder (from which you know that they are NOT identified automatically)
- Run the Organize-Task and then open the Organize-Dialog for the file that couldn't be processed automatically
- Adjust the settings and click OK
- The dialog will close and you can see that the file is being processed
- You don't need to wait for the copying to complete anymore. Instead you can immediately proceed with another file

### Test dialog error handling
- Put one or more files into the auto-organize folder (from which you know that they are NOT identified automatically)
- Run the Organize-Task and then open the Organize-Dialog for the file that couldn't be processed automatically
- Adjust the settings to a non-existing value (like season 123 and episode 987) and click OK
- The dialog will not close but display an immediate error message instead
- Thus, you can fix the problem and try to click OK again

### Concurrency
- Put one or more files into the auto-organize folder (from which you know that they are NOT identified automatically)
- Run the Organize-Task and then open the Organize-Dialog for the file that couldn't be processed automatically
- Adjust the settings and click OK
- At the same time (while the first request is being processed), try process the same file from another browser instance (or the same).
- The second operation will be blocked (also, manual and ScheduledTask processing can't simultaneously process the same file anymore)

softworkz:
Just updated with the desired modifications!

PR #2089, opened 2016-08-22
AutoorganizeLog: Fixed error when log empty; re-introduced full reload on task completion

softworkz:
Also removed unused method in ScheduledTaskWorker


softworkz:
Updated..

PR #2090, opened 2016-08-23
Core server preparation for new "TV Maze Provider Plugin"

softworkz:


softworkz:
Even if we move this to a plugin for now, it would be better to have a fixed enum for this in the server core..

PR #2091, opened 2016-08-23
Provide (unique) ETag value for synced media sources

softworkz:
The item.Id is valid for this purpose because it provides a unique identifier for the synced file.

Background: A persistent ETag value enables range-based streaming and seeking of static files being served via http.


LukePulverenti:
This isn't valid because it should be something that pertains to the file, not the database item.

softworkz:
In this case, "item" is of type "LocalItem", so item.Id is populated by "MediaSync.GetLocalId" and this depends on the sync job id.
Thus, I assumed that the Id is unique...

PR #2094, opened 2016-08-24
Encrypt Password when sent over the web

jose-pr:
Add the option to encrypt the password when authenticating with the server, in preparation of adding external authentication methods to decrypt the password and pass on to external authtentication services


jose-pr:
@LukePulverenti  Just added a secure service to the api, which i will need to send passwords that i can decrypt back, can also be use to send data to any other route in the server

LukePulverenti:
Ok just fyi we already have a home-grown text! require module. See if that can be used.

Make sure to use bower instead of thirdparty if possible.

Also, will with these changes, will Common be able to compile if re-targeted as a .net core project?

jose-pr:
Didn't know about the homegrown text. Is it already installed? Do i use it the same way as i am doing now? 
About the Common http://stackoverflow.com/questions/35363358/computing-sha1-with-asp-net-core it it possible as a nuget package

jose-pr:
@LukePulverenti done. Any other comment/concerns?

LukePulverenti:
So the core server would handle decryption and send the unencrypted version to external authentication providers?

LukePulverenti:
Also, is this pull request obsolete due to #2139? It appears to include some of the same changes. Thanks.

jose-pr:
Yes, the core server would have the option to receive an unhashed password encrypted , decrypt it and use for authentication.
 And yes the other pull request is based on this one

PR #2096, opened 2016-08-24
Create common commit and read methods for the repositories

jose-pr:


LukePulverenti:
Looks nice. How would this handle multiple commands in a write transaction? Or multiple result sets when reading?

LukePulverenti:
If you're looking at this in order to prepare for a different database, since everything is based on .NET interfaces I think all we need to do is supply an alternate implementation to create a different type of connection.

jose-pr:
In the future after finishing the external auth providers i was thinking more of using Entity Framework Core for all repositories which will create the database sqlite, mysql, postsql and tables and add columns by itself which works with 4.6.2 the current framework for now i am doing it because i was addingand reading some fields to the users database and was just rewriting the code multiple times, so i just put it in the base class. 
At the moment each commit, read is for one command is just to make it easier by handling the errors and the parameters without having a connection. Just do multiple commits or multiple reads, the commits does use writelock so if multiple writes are made it will wait.

PR #2105, opened 2016-08-26
Programmatic provider id handling for xml and nfo persistence

softworkz:


softworkz:
Updated..

softworkz:
Rebased.

MediaBrowser.Controller/Providers/BaseItemXmlParser.cs
```diff
@@ -60,6 +64,22 @@ public void Fetch(MetadataResult<T> item, string metadataFile, CancellationToken
                 ValidationType = ValidationType.None
             };
 
+            _validProviderIds = _validProviderIds = new Dictionary<string, string>(StringComparer.InvariantCultureIgnoreCase);
+
+            var idInfos = ProviderManager.GetExternalIdInfos(item.Item);
```
LukePulverenti, line 69: Is this a filter?

softworkz: Well, this just returns all provider ids (= classes implementing IExternalId) that apply to the specific item type (= IExternalId.Supports(item) returns true)
This is the exact same list of ids that are shown in the metadata editor form.

Let's compare the effect of this change to the previous implementation:
1. We are only importing metadata ids that Emby can understand and process for a certain item. (The previous implementation would have imported a "MusicBrainzAlbumId" even for a TV series episode, but there's no reason to import that for a case where the id doesn't apply anyway). I also checked that there is an IExternalId defined for each of the previously hardcoded id strings. (except for TvRage, but that's history)
2. We are now importing ALL provider ids that apply to the specific item, not only the hardcoded ones. (this even includes provider ids specified in plugins)

LukePulverenti: Right, I see, because without that list we don't know what to import, right?

softworkz: Correct :-)

PR #2114, opened 2016-08-28
QuickSeek Feature

softworkz:
Here comes the "QuickSeek" feature for Emby!
## Summary

The QuickSeek feature allows fast seeking in videos served by Emby that are remuxed or transcoded on-the-fly
## Instructions

To get this (experimental feature) to work, a number of specific requirements have to be met:
- A recent version of ffmpeg must be used and configured (I'm using 20160815)
- The feature must be activated in ServerConfiguration (config/system.xml); <EnableQuickSeek>true</EnableQuickSeek>
- Client browser must be Chrome connecting to the default web interface
- Video container mus be mkv
- Remuxing or transcoding must be used

### DIFF
--- a/MediaBrowser.Api/ApiEntryPoint.cs
+++ b/MediaBrowser.Api/ApiEntryPoint.cs
@@ -707,9 +707,12 @@ namespace MediaBrowser.Api
         public DateTime LastPingDate { get; set; }
         public int PingTimeout { get; set; }
 
+        public DateTime CreatedAtUtc { get; private set; }
+
         public TranscodingJob(ILogger logger)
         {
             Logger = logger;
+            CreatedAtUtc = DateTime.UtcNow;
         }
 
         public void StopKillTimer()
--- a/MediaBrowser.Api/MediaBrowser.Api.csproj
+++ b/MediaBrowser.Api/MediaBrowser.Api.csproj
@@ -59,6 +59,7 @@
     <Reference Include="System.Core" />
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Data" />
+    <Reference Include="System.Net" />
     <Reference Include="System.Xml" />
     <Reference Include="ServiceStack.Interfaces">
       <HintPath>..\ThirdParty\ServiceStack\ServiceStack.Interfaces.dll</HintPath>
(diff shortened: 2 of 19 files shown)


softworkz:
## Benefits
The primary benefit is (nearly) instant seeking within a short range of the current playback position.
Primary use cases are things like skipping commercials or repeating scenes.
Depending on the client't browser cache settings, it's also possible to repeat a long sequence without additional network traffic.

## Limitations
This feature doesn't help when skipping forward large timespans. In these cases it's difficult for html5 clients to determine the correct byte range to request on one side, and on the other side, the transcoding process might not yet have reached the skip destination.
In such cases, we are setting up a new transcoding session just like usual.

softworkz:
PS: To demonstrate the actual benefit and effect of this feature, I added two new "Skip" buttons to the video player. The left button skips back 5s and the right button skips forward 20s.

softworkz:
A few more remarks;

In order not to break any clients with this feature, the only viable solution seemed to me that a client must explicitly indicate that it can handle the feature. In our case that's these two aspects:
- The media file has a duration (although it's still being encoded/remuxed)
- The media is served via http range requests/responses

So I implemented the following procedure:
- When the client player calls "getPlaybackInfo" it can add a new parameter "CanQuickSeek"
- The server checks if all conditions for QuickSeek are met. If yes, it returns a streaming url containing the parameter "tryquickseek=true"
- When the client requests a stream and the stream url contains "tryquickseek=true", the server re-evaluates all conditions and will then finally set "StreamState.UsingQuickSeek"
- If the latter is true, the file will be served with a preliminary duration in the header and with support for range requests

LukePulverenti:
FYI, chrome transcoding has been switched to HLS, therefore this is obsolete now as the the HLS protocol has support for these kinds of things already. This probably should just be closed.

softworkz:
A while ago I asked why we are using the mkv container for streaming anyway.
The argument (which I found reasonable) was: that sticking to MKV would allow the server to remux only instead of transcoding.
I don't know much about HLS and if it can be used with remuxing.
But if you say that in the future we will no longer have any remuxing or transcoding to MKV, then we can drop that code indeed...

LukePulverenti:
hls has it's own set of complexities and drawbacks but other new features have made it a necessity.

softworkz:
I'm fine with it. MKV has never been a good option for streaming anyway. 
I just wish we had come to that conclusion a little earlier.. :-)

softworkz:
(I really pushed those ffmpeg guys a little to make them like my changes)

PR #2125, opened 2016-08-29
QuickSeek Feature (server side only)

softworkz:
This is like PR #2114 but only including the server-side changes


LukePulverenti:
FYI, chrome transcoding has been switched to HLS, therefore this is obsolete now as the the HLS protocol has support for these kinds of things already. This probably should just be closed.

PR #2128, opened 2016-08-30
Small display adjustments in ItemDetails screen

anderbytes:
I'm bothered by how small some importants things are shown in ItemDetails screen.

My changes here try to make this screen a little more similar to seen in Emby Theater or Emby TV.

If the way I did the changes aren't the correct way of doing this... please consider correcting it instead of discarding the whole idea.

Thanks.

### DIFF
--- a/MediaBrowser.WebDashboard/dashboard-ui/css/librarybrowser.css
+++ b/MediaBrowser.WebDashboard/dashboard-ui/css/librarybrowser.css
@@ -216,6 +216,7 @@
 
 .itemOverview {
     white-space: pre-wrap;
+    font-size: 20px;
 }
 
 a.itemTag:hover {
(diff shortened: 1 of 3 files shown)


MediaBrowser.WebDashboard/dashboard-ui/css/librarybrowser.css
```diff
@@ -216,6 +216,7 @@
 
 .itemOverview {
     white-space: pre-wrap;
+    font-size: 20px;
```
LukePulverenti, line 219: We generally try to avoid fixed font sizes, and instead just let the sizes flow from the top of the document down.

anderbytes: I understand. But still, this may need adjust, I believe.
I consider it a little too small, and the style doesn't guide it's sizing at all, today.

anderbytes: Hi Luke. Is this pull going somewhere?

PR #2164, opened 2016-09-13
Service restart and update

hatharry:


LukePulverenti:
Thanks. Can you explain the approach? What is CanRestartWindowsService querying for?

hatharry:
When emby is started it checks to see if Service Control (sc.exe) can be executed by the current service user and has the correct permissions to query the service. When this is true the restart button appears in the web UI, as we know the service user can restart the service using Service Control (sc.exe).

On restart the app is disposed then Service Control (sc.exe) is excuted by cmd to stop and start the service.

LukePulverenti:
Hi, can you look at this?
http://emby.media/community/index.php?/topic/39243-service-wont-restart-after-upgrade
Thanks.

PR #2245, opened 2016-10-19
Service restart fix

hatharry:


LukePulverenti:
Thanks but this is all beginning to get hacky. It is beyond my level of comfort. At this point I would rather rewrite the entire windows service integration to be based on this
https://github.com/kohsuke/winsw

PR #2276, opened 2016-11-08
Remove zerolatency from ffmpeg h264.

x-cimo:
Even if zerolatency sounds good for streaming video, it's purpose is real time video such as video conference, webcam etc.

zerolarency greatly reduce video encoding performance (by as much as a factor of 2), hurting picture quality and compression by limiting how much processing can be done in favor of low latency.


LukePulverenti:
Done, thanks.

PR #2328, opened 2016-12-04
Add Service Wrapper

hatharry:
Hey,
Here is a wrapper for emby it works using exit codes. To use enter the following:
```
MediaBrowser.ServerApplication.exe -uninstallservice
Emby.Service.exe -installservice
```

Source for the service Wrapper can be found here [https://github.com/hatharry/Emby.Service](https://github.com/hatharry/Emby.Service)


hatharry:
The "Emby.Service.exe" executable launches Emby server as a task. This causes the service to enter the started state as fast as possible. The executable also restarts the server process without restarting the service. This is achieved by comparing exit codes from the server process and restarting the process if necessary. Automatic Updates also work due to the service being the same name. On service stop the process is also stopped. 

The Emby server itself has two options added "-norestart" which causes Emby to exit with a code instead of restarting. and "-notray" which removes the tray icon because it is not needed when running as a service.

PR #2359, opened 2016-12-20
Improved server sync event handling

softworkz:


softworkz:
Modified PR: Removed SyncJobUpdated server event

PR #2377, opened 2017-01-03
Hls streaming fixes

softworkz:


softworkz:
Wrong target branch..

PR #2395, opened 2017-01-13
Fix M3U Channel number parsing 

sle118:
The current m3u parsing wasn't taking into account that some OTA tuner channel numbers are decimal rather than integers. 

From wikipedia page [M3U](https://en.wikipedia.org/wiki/M3U)
The following are M3U extensions:
Directive	Description	Example
#EXTM3U	File header. Must be the first line of the file.	
#EXTM3U
#EXTINF	Track information, including runtime and title.	
#EXTINF:191,Artist Name - Track Title

The proposed PR uses character "-" as the channel number separator rather than using the dot character.


LukePulverenti:
Who created this m3u? what is the source?

LukePulverenti:
This would break this example:
#EXTINF:0,84. VOX Schweiz

LukePulverenti:
I whipped up a unit test, so just make sure these tests are passing, and add a test for the scenario you are trying to cover.

https://github.com/MediaBrowser/Emby/blob/dev/MediaBrowser.Tests/M3uParserTest.cs

Thanks.

LukePulverenti:
Nevermind, i took care of it. thanks !

sle118:
That was quick! Thank you! At least I feel like I was able to help with root cause analysis, which is better than nothing. 

The problem with m3u with regards to channels is that there doesn't seem to be a firm standard. Your code does "brute force analysis" and should work for 90% of the cases, making it more future proof.

Thank you!

PR #2443, opened 2017-02-05
Respect a given runtime for live tv and recordings

puenktchen:
If a tv backend provides this information already, than it's doing it for a good reason.
There's no reason to overwrite the runtime in such a case with one determined by ffprobe.
A particular use case is an inprogress recording with a file as input. This results in wrong durations in the players if ffprobe's RunTimeTicks are used.


LukePulverenti:
Is this still needed?

puenktchen:
I don't think so.

PR #2490, opened 2017-02-25
Removed superfluous notes from LICENSE.md

voxadam:
All text past line 278 of LICENSE.md is unnecessary and not actually part of the GNU General Public License v2. The text is only there to document how a developer should document that their code is subject to the GPL v2.


slang800:
Also, it might be worth renaming `LICENSE.md` to `LICENSE`, since the GPL wasn't written in Markdown.

ian-kelling:
> All text past line 278 of LICENSE.md is unnecessary and not actually part of the GNU General Public License v2. 

That is incorrect, per gnu.org:

https://www.gnu.org/licenses/gpl-faq.en.html#GPLOmitPreamble

> Can I omit the preamble of the GPL, or the instructions for how to use it on your own programs, to save space? 
> The preamble and instructions are integral parts of the GNU GPL and may not be omitted.

This may be closed. I do suggest renaming LICENSE.md to LICENSE or, as gnu suggets COPYING.

PR #2497, opened 2017-02-28
Legacy HDHomerun Tuner Support

KeyserSoze1:
**Background**
=================
I find myself needing to get  broadcast tv channels to where putting an antenna is not feasible but I have network connectivity.  I have an old HDHomerun and decided to make it a pet project to find a way to get it working with something.

**Requirements**
=================
A legacy HDHomerun device (I've currently only tested with a HDHR3-US) with the latest firmware installed.
Install the HDHomerun software on Windows.  This goes through the process of finding the legacy device and scanning and creating a json lineup that is stored on their side.  I do not believe this works on the Mac/Linux side yet.

**Caveats**
=================
I have tested this so far using a HDHR3-US with Emby running on windows and video streaming to a browser and a FireTV.  So far it has been working well.  I hope to test with Emby running on linux soon.

**Input Requested**
=================
Is there interest in adding this support to Emby?  If not, I'm not going to spend a lot of time cleaning up what I have done so far.

If there is I would like input on how to best approach networking changes.
Legacy HDHomeruns are controlled with commands sent over TCP.  It doesn't appear that making connections to an external TCP server (outside of http) has been done much with Emby.  I've made a hack in SocketFactory/UdpSocket to allow it to work but I'm looking for input on what would be considered the best approach for this.  Renaming IUdpSocket to ISocket would be the minimum since IUdpSocket does't really implement any UDP specific behavior.   Making an ITcpSocket class is another option.  There are probably others.

Any other input is appreciated.  My C# experience is somewhat limited.  I've done projects with it in the past but nothing complicated.  I started with the HDHomerun tuner classes already there and made modifications and additions as necessary.


LukePulverenti:
Hi, very cool. Is this for internal PC tuners or legacy network tuners? or both?

KeyserSoze1:
This is for legacy network tuners. The models that don't support Upnp.  They use special messages sent over tcp to interact with the device.  The video stream is sent over udp once the correct set of commands has been sent.

KeyserSoze1:
Pushed a few fixes so that it no longer works just for US broadcast channels but now should work for all channel types.

LukePulverenti:
Thanks. If you can please sign the CLA then I will look at integrating this very soon.

KeyserSoze1:
Pushed the last of my changes up and signed the CLA.  Let me know if you have any questions.

LukePulverenti:
Merged manually, thanks !

PR #2509, opened 2017-03-05
Fix naming of nvenc codecs. It should be nvenc_h264.

gerrit507:


gerrit507:
It seems like both nvenc_h264 and h264_nvenc is used, but nvidia recommends h264_nvenc now https://developer.nvidia.com/ffmpeg
Just keep that in mind if an older ffmpeg version is used.

PR #2550, opened 2017-03-28
MediaBrowser: Return null from GetInputFormat when in mts format.

howlett:
ffmpeg does not have an '-f mts' option.  When returning mts from
GetInputFormat, that option is appended to the command arguments for
transcoding and thus ffmpeg fails with:

Unknown input format: 'MTS'

Signed-off-by: Liam R. Howlett <howlett@gmail.com>

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -184,6 +184,10 @@ namespace MediaBrowser.Controller.MediaEncoding
             {
                 return null;
             }
+            if (string.Equals(container, "mts", StringComparison.OrdinalIgnoreCase))
+            {
+                return null;
+            }
 
             return container;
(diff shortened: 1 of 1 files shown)


LukePulverenti:
Can you submit on the dev branch? Thanks !

LukePulverenti:
I'm working towards getting a release out so i've taken care of it. thanks.

howlett:
This patch still works on the dev branch and is still needed but it appears you have rewritten the encoding portion which does not detect this stream as AVC, which it is.  And so the copy ffmpeg codec is used which is wrong.  Even though I was able to shortcut this to use the older encoding it still doesn't play on my phone.

LukePulverenti:
Nothing related to detection has changed. Best thing to do is provide the info requested in how to report a media playback issue: https://emby.media/community/index.php?/topic/739-how-to-report-a-problem/

Thanks !

PR #2581, opened 2017-04-16
Remove link to non-functional API documentation

voxadam:
Until integrated API documentation (e.g. Swagger) has been restored this link should be removed as the 404 type result is disconcerting and gives the impression of a less polished product. This link, or something similar to it, can be restored once the documentation has been reimplemented.


LukePulverenti:
swagger will (begin to) return for the next release, thanks.

PR #2587, opened 2017-04-20
fix comment in Dockerfile

slang800:
that's not Dropbox you're downloading from


LukePulverenti:
Thanks ! I'll handle this on the dev branch.

PR #2670, opened 2017-05-29
Fixed ResultFactory not being set in Plugins

MrSmoke:
As per the doc https://github.com/MediaBrowser/Emby/wiki/Creating-Api-Endpoints if using the interface `IHasResultFactory`, ResultFactory should be automatically set. 

`IHasResultFactory` should really just be passed in the ctor (which also works), but then the documentation should reflect that.


LukePulverenti:
The core endpoints use the same method as plugins, this shouldn't be necessary. There's got to be some other factor here that's going on.

MrSmoke:
That's what I thought, but there's no classes in the project that implement `IHasResultFactory` that I could find (or at least according to Visual Studio). The `BaseApiService` seems to just use `ApiEntryPoint.Instance.ResultFactory`.

LukePulverenti:
Yes I think going forward it is best to just pull it in via constructor rather than the magic property that we used to set before. Thanks.

PR #2688, opened 2017-06-03
Cleanup

Bond-009:
Fixed some warnings
Removed unneeded ToLists
And more


LukePulverenti:
This isn't safe to merge. I just eyeballed it quickly and I can give you two examples. Many of the ToList() are not needed, specifically look at the ones that are happening inside of a lock statement. By removing the ToList(), the enumeration of the list is now moved to outside of the lock. Audio.cs - without the ToList() you will end up modifying the original list, which is not intended.

Bond-009:
Sorry, I will take a look at what this brakes tomorrow.

PR #2719, opened 2017-06-21
Fix an issue causing LiveTV to stop working if a connection to a hdho…

x-cimo:
Fix an issue causing LiveTV to stop working if a connection to a hdhomerun was lost.  There was no handling of connection lost causing "dead" stream to remain in the LiveStream list. Client trying to start the same channel would find the stream but ffprobe would hang because there is no data coming over the disconnected connection.

This set the connection as disconnected, and futher attempt to start a channel would first cleanup the livestream list and only use "working" tv tuner.

Note: I tried hard to find a callback path directly from the exception handling in HdHomerunHttpStream to get a pointer to the EmbyTV class, in order to remove the tuner from the _liveStream right away, but because of the way the code is structured, I had to set a flag on the tunerhost that is later checked by the EmbyTV class.  

Tested and resolve my issue with hdhomerun tuners (and Emby getting completly stuck).

### DIFF
--- a/Emby.Server.Implementations/LiveTv/EmbyTV/EmbyTV.cs
+++ b/Emby.Server.Implementations/LiveTv/EmbyTV/EmbyTV.cs
@@ -1139,6 +1139,20 @@ namespace Emby.Server.Implementations.LiveTv.EmbyTV
 
             try
             {
+                // Remove all tuner connection which are no longer connected
+                // Iterate backward so we can modify the list as we iterate
+                for (int i = _liveStreams.Count - 1; i >= 0; i--)
+                {
+                    if (_liveStreams[i].isDisconnected())
+                    {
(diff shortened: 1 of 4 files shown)


x-cimo:
Any update on getting this merged or a different fix being developed?

LukePulverenti:
I'm not convinced this is necessary because there is already a loop in that class that will handle disconnection. How can i simulate this for testing? Because i started watching a channel in the web app, then pulled the network cable out of the HD Homerun. As expected this caused playback to fail, and so I went back to the channel detail screen. Then I plugged the network connection back in, gave it a few seconds and I was able to click play and start playing again.

LukePulverenti:
The code has changed so this is no longer applicable. I am happy to review if you can prove that it is needed, but there are already checks in place to handle dropped connections. Thanks.

PR #2957, opened 2017-10-15
Update de.json

DaniGTA:
Altiviere not exist in german.
The correct spelling is Aktiviere.


LukePulverenti:
Hi, please sign up at onesky and contribute to the Media Browser project. Your changes here will just get overwritten anytime we pull the latest translations. Thanks.

PR #3027, opened 2017-11-19
Fix skip of install vcredist 2015 if 2017 is installed

AeonLucid:
The vcredist 2015 can not be installed if vcredist 2017 is already installed, therefore it should not try to install.

This PR checks if vcredist 2017 is installed and if so, skips the install.

### DIFF
--- a/MediaBrowser.ServerApplication/MainStartup.cs
+++ b/MediaBrowser.ServerApplication/MainStartup.cs
@@ -666,8 +666,16 @@ namespace MediaBrowser.ServerApplication
 
                 if (Environment.Is64BitProcess)
                 {
+                    // Visual C++ 2015 64bit
                     key = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Default)
                        .OpenSubKey("SOFTWARE\\Classes\\Installer\\Dependencies\\{d992c12e-cab2-426f-bde3-fb8c53950b0d}");
+
+                    // Visual C++ 2017 64bit
+                    if (key == null)
+                    {
+                        key = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Default)
+                            .OpenSubKey("SOFTWARE\\Classes\\Installer\\Dependencies\\,,amd64,14.0,bundle");
+                    }
 
                     if (key == null)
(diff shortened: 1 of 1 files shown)


softworkz:
Thanks a lot for your effort to help us out in that matter.

Although I didn't try, I believe that  it could have worked as a temporary fix.

But we need to think a bit further and from larger perspective, it's not a durable and reliable solution (1). Also it doesn't directly address the actual source of the problem (2).
 

Let's look at aspect (1) first:

http://stackoverflow.com/questions/12206314/detect-if-visual-c-redistributable-for-visual-studio-2012-is-installed

The previous code was following the reference above. But this does not describe the optimal way. 

 Depending on Windows Installer and dependencies is not too reliable since the installer keys may change with newer versions.

I don't even know if those keys would be the same when someone includes them into his setup via the merge modules.

To understand this better, there's one thing you need to know:

For the first time, VC libs remain compatible between two versions of Visual Studio
No need to install the VS2015 runtime. The 2017 version will cover both.
(this does not apply to any other two VS versions)
<https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017>

 

With this situation, we don't even know how long this compatibility will be kept. A Visual Studio 2019 could include a new  runtime that is still compatible, but it will be delivered with an installer having all different IDs internally. That's why it's not the best idea to do the detection via installer IDs.

 

 

Now I am using the same detection mechanism as the vcredist installer packages themselves:
https://docs.microsoft.com/en-us/cpp/ide/deploying-visual-cpp-application-by-using-the-vcpp-redistributable-package

This helps to avoid discrepanices in detection of installation status. (and it is also the MS recommended way).

About the actual problem that this PR didn't address (2):

There's a bug in some of the VS 2017 redist installers in that they are not updating their version information to both registry locations (i.e. x64 and x86), just in one of them (one with WOW6432) and one wihout)

 That's why I'm reading both values now and take the higher build number to compare.

 Also I changed everything to use the VC 2017 redist.(since it's is just a slightly newer version of the 2015 dlls).

softworkz:
Thanks again for your help and identifying the problem.
I'm closing.

AeonLucid:
Gotcha :)

PR #3057, opened 2017-12-05
Enable deinterlacing for transcoding with vaapi.

gerrit507:
This if clause should enable deinterlacing while transcoding with vaapi. This is very useful if Live-TV is used. At the moment the channels are just streamed interlaced if vaapi is enabled.


LukePulverenti:
Looks like this pull request has 0 changes?

gerrit507:
Sorry I've created a new one. This one had a unnecessary if clause in it.

PR #3058, opened 2017-12-05
Enable deinterlacing for transcoding with vaapi

gerrit507:
This if clause should enable deinterlacing while transcoding with vaapi. This is very useful if Live-TV is used. At the moment the channels are just streamed interlaced if vaapi is enabled.


LukePulverenti:
Thanks !

gerrit507:
@LukePulverenti Thank you for the new build. It works and the quality of Live-TV looks great.

PR #3185, opened 2018-03-02
+ workaround for tvheadend channel icons

gerrit507:
Set the contentType to image/png when tvheadend doesn't provide any contentType.

### DIFF
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -161,6 +161,15 @@ namespace MediaBrowser.Providers.Manager
 
             }).ConfigureAwait(false))
             {
+                //workaround for tvheadend channel icons
+                if (response.ContentType == "")
+                {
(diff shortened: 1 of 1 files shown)


LukePulverenti:
shoot, i merged this by accident thinking this was another PR for the tvh plugin. this is a little too dirty for my liking.

gerrit507:
I know it is dirty :/ but I didn't found any other way.

gerrit507:
I'm now working on a binary reader do deal with the issue. Looks promising.

LukePulverenti:
I think what you're going to need in addition to that is a custom api added to the plugin that will serve as an in-between. And then you can have the plugin return a localhost url for the image, and when the api gets hit, it will download the image from tvheadend and figure out the content type so that it can fill out the response headers.

gerrit507:
Yes it would probably be the best to do it in the plugin. I've implemented a reader in the emby-server for now. If tested it a couple of times and it works. the advantage of this is that it can also handle missing mimeTypes of other plugins.
#3187

PR #3187, opened 2018-03-02
binary reader for finding file types of images

gerrit507:
This reads the so called magic number from a file and sets the mimeType accordingly.

See this for reference:
http://www.c-sharpcorner.com/UploadFile/a8024d/how-to-find-file-type-from-magic-number-of-file-in-mvc/
https://en.wikipedia.org/wiki/List_of_file_signatures


gerrit507:
@LukePulverenti 
Don't you prefer this approach over  #3185 ?

LukePulverenti:
This method will be slower so i will just use your previous hack just so that we can get the images working. thanks.

PR #3227, opened 2018-04-06
Addressed performance issue when dealing with large EPGs. 

mak0t0san:
When refreshing the guide, a List<ChannelInfo> would get built and then queried using .FirstOrDefault(), which is slow with large lists.  

A dictionary lookup is significantly quicker, with a small trade off being that we create 3 dictionaries to act like indexes for our data, one based on ChannelInfo.Id, another based on ChannelInfo.Name, and finally ChannelInfo.Number.  If you have a large M3U playlist (8000 or so entries), as well as large EPG data (14+ MB), the call to FirstOrDefault would get called repeatedly which causes the list to be scanned multiple times, along with calling a function for every time to format the string, or performing a case insensitive search.

To get around the case insensitivity with the dictionary, all the keys have been made to UPPERCASE so that it will match.

There's still room for improvement but now the guide will refresh in a more reasonable time.  Before this fix, I left it running overnight and it was only 16% complete.


LukePulverenti:
Thanks. I wrote this slightly different to the way you did it, although conceptually the same idea. Thanks.

PR #3355, opened 2018-07-10
Bugfix: Incorrect use of DateTime.ToUniversalTime() results in incorrect dates returned by metadata providers

softworkz:
Accidentially I've seen that this issue is still unresolved.

Here's an updated version of my 2years old PR. I still think that this is the right (small) solution unless we want to introduce default Read- and Write-Configurations (== time zone +  datetime format) for all providers

https://emby.media/community/index.php?/topic/26050-premiered-date-is-showing-up-wrong/page-7#entry595893

Here's the original description:

This is a revised approach to fix the problems caused by incorrect
timezone adjustments.

- Let's look at MovieDbProvider, Line 97
- DateTime.TryParse creates a date from the string in the xml

- The next statement ".ToUniversalTime()" converts the datetime from
local a local datetime to a UTC datetime

- Emby-Server1 is in Timezone UTC+3: If the parsed date is "2016-03-30
01:00" the converted date would be "2016-03-29 22:00"
- Emby-Server2 is in Timezone UTC-9: If the parsed date is "2016-03-30
01:00" the converted date would be "2016-03-30 10:00"

As a result, Emby-Server1 would store a different item date than
Emby-Server2. This is incorrect because the UTC date (per definition)
must be a constant value that is independent of any time zone.

The previous pull request suggested to remove all UTC conversions but
was turned down because it would break consistency of the API.
This commit follows a different approach: It assumes a constant timezone
(US EST - Eastern Standard Time) for all dates returned from the
affected metadata providers.
The result: Emby will continue to use UTC based dates, but the
conversion is done consistently.

### DIFF
--- a/Emby.Server.Implementations/HttpServer/HttpResultFactory.cs
+++ b/Emby.Server.Implementations/HttpServer/HttpResultFactory.cs
@@ -702,7 +702,7 @@ namespace Emby.Server.Implementations.HttpServer
 
                 if (DateTime.TryParse(ifModifiedSinceHeader, out ifModifiedSince))
                 {
-                    if (IsNotModified(ifModifiedSince.ToUniversalTime(), cacheDuration, lastDateModified))
+                    if (IsNotModified(ifModifiedSince, cacheDuration, lastDateModified))
                     {
                         return true;
                     }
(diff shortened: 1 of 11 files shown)


LukePulverenti:
No longer applicable.

PR #3404, opened 2018-08-14
Fix german translation for 'remember me'

lexxodus:
'Remember me' on the login page was translated in german to 'Erinnere mich' which actually means 'Remind me'. The literal translation for this would be 'Erinnere dich an mich' which would be a bad expression given this context. In german usually the frase 'Angemeldet bleiben' is used, which literally means 'Stay logged in'.


LukePulverenti:
Hi there, thanks for the contribution! Can you please submit this text on Onesky instead? Any changes you make here will just get replaced whenever we update the text from Onesky. 

To update on onesky, simply sign up on their website, and then search for the MediaBrowser project. Thanks !

PR #3645, opened 2020-10-18
tell people where to find new versions of Emby.

foxt:


zero77:
Why not reference this also.
https://github.com/MediaBrowser/Emby.Releases/releases

PR #3830, opened 2024-03-27
Create azure-webapps-node.yml

xuhao61:


softworkz:
Such PRs are exploitation attempts. We have disabled GH action runs on public repo PRs, so I can just say: Nice try, but you failed!

